PR #13, opened 2025-03-13
Implement pext bitboards for rooks

clarashepherd:
Unit tests passing both with BMI2 enabled and disabled.


src/movegen/rooks.h
```diff
@@ -6,6 +6,9 @@
 #include <array>
 #include <cstdint>
 
+#include "magics/hashing.h"
+#include "magics/pext.h"
```
clarashepherd, line 10: I tried conditionally including the headers, but caused issues with below `...pext` and `...hashing` functions.

Is there a nicer pattern here?

hansbinderup (commented):
Some initial comments! 
Looks good, very happy to have you onboard :100:

src/movegen/rooks.h
clarashepherd: Done. `getRookMovesHashing` would also have to be wrapped in `#ifdef` logic, so removed this redundant function naming and implemented conditional in `getRookMoves` directly.

Same for `generateRookAttackTable`.

Is this an acceptable solution? We're still avoiding nasty inlining of the `#ifdef`.

meson.build
```diff
@@ -9,6 +9,14 @@ subdir('version')
 add_project_arguments('-fconstexpr-loop-limit=2147483647', language: 'cpp')
 add_project_arguments('-fconstexpr-ops-limit=100000000000', language: 'cpp')
 
+cc = meson.get_compiler('cpp')
+bmi2_supported = cc.get_supported_arguments('-mbmi2').length() == 1
```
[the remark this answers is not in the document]
clarashepherd: I've added the option and set it to false by default. Or would you prefer to default to true?

We should also decide on what options are in the scripts, and consider checking both options in the nightly build - some bugs may only be visible with the flag either true or false.

src/magics/pext.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <cstdint>
+#include <immintrin.h>
+
+namespace magics {
+namespace pext {
+
+// Efficient dense indexing of blockable pieces
+// https://www.chessprogramming.org/BMI2#PEXT_Bitboards
+constexpr inline int magicIndex(uint64_t occupancy, uint64_t mask, bool is_compile_time = false)
```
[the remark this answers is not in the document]
clarashepherd: That makes sense, thanks for the spot. This function has now been removed anyway :)

src/magics/pext.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <cstdint>
+#include <immintrin.h>
+
+namespace magics {
+namespace pext {
```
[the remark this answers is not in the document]
clarashepherd: Noice. Implemented for the hashing, `pext.h` ended up empty so I've removed it for now. We can reinstate it we need to add complexity to the index calculation, for some reason.

meson.build
```diff
@@ -17,6 +17,10 @@ if get_option('buildtype') != 'debug'
   add_project_arguments('-funroll-loops', '-m64', '-finline-functions',  language : 'cpp')
 endif
 
+if get_option('use-bmi2')
+    add_project_arguments('-mbmi2', language: 'cpp')
+    add_project_arguments('-DUSE_BMI2', language: 'cpp')
+endif
```
hansbinderup, line 23: We talked about adding cross compilation targets. I think that's a better way to distinguish if it should be enabled. 

Instead of adding our own compiler flag we can use the native flag based on CPU/compilation target:

`-march=x86-64-v2` -> bmi not enabled due to arch
`-march=x86-64-v3` -> bmi enabled due to arch

We can then test for BMI2 availability (something like):

```cpp
#ifdef __BMI2__
  #define PEXT_MAGICS 1
#endif
```

.. and based on that decide which direction to take.
It will mean that anyone compiling Meltdown with `-march=native` will automatically compile the best variant for their cpu :)

And for our releases, we don't have to add our own flags but instead use what's available from the compiler.

NOTE: if we only use the BMI2 check in one place then we might as well just use that flag directly so we're not making a dependency chain for no reason :)

src/magics/hashing.h
```diff
@@ -0,0 +1,95 @@
+
+#pragma once
+
+#include <array>
+#include <cstdint>
+
+namespace magic::hashing {
```
hansbinderup, line 7: Nitpick: We will probably add bishops here as well? Then we might want to namespace instead of prefix
```suggestion
namespace magic::hashing::rooks {
```

src/movegen/rooks.h
```diff
@@ -80,6 +86,7 @@ constexpr uint64_t rookAttacksWithBlock(int square, uint64_t block)
     return attacks;
 }
 
+#ifdef USE_BMI2
```
hansbinderup, line 89: I think that it would be cleaner if we could use as few of these checks in this file and move as much of the specifics into header files that we then include.

So eg:

```cpp
#ifdef PEXT_AVAILABLE
  #include "pext_magics.h" // or what name you prefer
#else
  #include "hashing_magics.h" // same here
#endif 
```

src/movegen/rooks.h
```diff
@@ -92,27 +99,65 @@ constexpr auto generateRookAttackTable()
...
+// https://www.chessprogramming.org/BMI2#PEXT_Bitboards
+static inline uint64_t getRookMoves(BoardPosition pos, uint64_t occupancy)
+{
+    occupancy = _pext_u64(occupancy, s_rookMasksTable[pos]);
+
+    return s_rookAttackTable[pos][occupancy];
+}
+#else
 /* https://www.chessprogramming.org/Magic_Bitboards */
 static inline uint64_t getRookMoves(BoardPosition pos, uint64_t occupancy)
```
hansbinderup, line 153: These methods have the same signature, so I thought we might as well move them to their respective headers and simply use "rooks.h" to include the methods that we're currently compiling for?
Then there'll basically be nothing in this file but merely the includes meaning that it will be completely agnostic from the consumers of the interface :)

hansbinderup (commented):
Could have been more clear about the direction before, sorry! I tried to be more specific in my comments here.
Also spent some more time looking into compiler flags and thought that this might be a very clean and scalable solution.
Let me know what you think :)

src/movegen/rooks.h
```diff
@@ -102,17 +119,20 @@ constexpr auto generateRookAttackTable()
 }
 
 constexpr auto s_rookAttackTable = generateRookAttackTable();
+} // end namespace
 
-}
-
-/* https://www.chessprogramming.org/Magic_Bitboards */
 static inline uint64_t getRookMoves(BoardPosition pos, uint64_t occupancy)
```
hansbinderup, line 124: This would probably be a small gain as well. (feel free to change if you like)
```suggestion
constexpr uint64_t getRookMoves(BoardPosition pos, uint64_t occupancy)
```

hansbinderup (approved):
Feel free to update the namespaces if you like (as commented).

Otherwise LGTM! Remember to squash the commits or perform a squash merge :)

PR #14, opened 2025-03-13
workflows: implement nightly build/release

hansbinderup:
This commit adds a script to build and pack a release. Also it adds a github workflow to trigger a nightly build; both scheduled automatically but also from a workflow dispatch.

The nightly releases will be tagged as prerelease drafts meaning that they will not polute the "real" releases.

NOTE: this will add the option to decide which branch to build a nightly from. This is a design choice as it will be quite handy to build a nightly from a feature branch etc. Also, it will allow the auto deploy (WIP) to fallback to "regular nightly" after a given period; perfect for testing!

TODO: easy way to turn it on/off if we want a more longterm test done eg.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,20 +1,51 @@
 name: Nightly
 
+# NOTE: this job should only build a release if either:
+# 1. triggered manually (workflow_dispatch)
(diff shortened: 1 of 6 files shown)


hansbinderup:
@clarashepherd no need to review - just tagged you to inform you about the new "feature"

PR #17, opened 2025-03-16
workflows: add create_release

hansbinderup:
This workflow will be dispatched to create new releases. An input will contain version and if the release should be a pre-release.


hansbinderup:
@clarashepherd gonna merge this one (can't test it before the workflow is on main), so let's hope it works :)

PR #20, opened 2025-03-16
move_gen: Implement pext bitboards for rook

clarashepherd:
Almost identical to rook case.


src/movegen/bishops.h
```diff
@@ -115,12 +132,16 @@ constexpr auto s_bishopAttackTable = generateBishopAttackTable();
 /* https://www.chessprogramming.org/Magic_Bitboards */
```
hansbinderup, line 132: nitpick: comment is already at the hashing part

hansbinderup (approved):
Sweet :+1:

PR #21, opened 2025-03-16
static_evaluation: improve tapered score + add some new positional bonuses / game phase bonuses

hansbinderup:
See each commit for details.

Let me know if the whole game phase stuff is too twisted @clarashepherd. 
I thought it's nice that all our scoring functions consider both mg and eg. Then it's easier to assign an actual meaningful score and do that consistently.

### DIFF
--- a/src/evaluation/positioning.h
+++ b/src/evaluation/positioning.h
@@ -5,6 +5,8 @@
 #include <array>
 #include <cstdint>
 
+#include "evaluation/game_phase.h"
+#include "evaluation/pesto_tables.h"
 #include "evaluation/position_tables.h"
 #include "helpers/bit_operations.h"
 #include "movegen/bishops.h"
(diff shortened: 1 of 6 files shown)


src/evaluation/positioning.h
```diff
@@ -15,126 +17,224 @@ namespace evaluation {
...
+constexpr std::array<int32_t, magic_enum::enum_count<gamephase::Phases>()> s_queenMobilityScore { 1, 1 };
+constexpr std::array<int32_t, magic_enum::enum_count<gamephase::Phases>()> s_kingShieldScore { 5, 5 };
 
 }
 
 namespace position {
 
 template<Player player>
-constexpr static inline int32_t getPawnScore(const BitBoard& board, const uint64_t pawns)
+constexpr static inline gamephase::Score getPawnScore(const BitBoard& board, const uint64_t pawns)
```
clarashepherd, line 47: My naive profiling (move evaluation from kiwiPete position) suggests that calls to `getPawnScore` already use a similar order of magnitude of CPU time to evaluation of rook or bishop moves. 

Screenshot from 2025-03-17 12-18-38

I think we'd need to justify the extra operations for separate mg and eg scores here with a performance gain from tweaking the parameters.

Or am I missing something?

clarashepherd (commented):
Perhaps we could get a clearer idea of the performance benefits vs drawbacks of this approach before going ahead and implementing it?

hansbinderup:
> Perhaps we could get a clearer idea of the performance benefits vs drawbacks of this approach before going ahead and implementing it?

I agree! I'm gonna close this PR for now. It seems to be quite a lot worse than what we have now. I'll think about a better way to implement a more scaleable tapered eval :+1: Thanks for the reviews :)

PR #22, opened 2025-03-19
tests: add cutechess tournament tests

hansbinderup:
This commit adds a new way of testing the engine.
It will run games with a given config and evaluate the performance between the two engines.

It is possible to setup the amount of games, if the test should run SPRT etc.
More info is added to the README.md.


clarashepherd (approved):
Looks good, approved.

PR #23, opened 2025-03-19
move_ordering: use primitive insertion sort

hansbinderup:
Previously std::stable_sort was used as the sorting algortihm. This adds quite a bit of overhead for sorts of our size.

Instead implement a primitive insertion sort which is also deemed fully deterministic + low runtime overhead.

Test results:

Score of Meltdown Build vs Meltdown v1.1: 113 - 52 - 120  [0.607] 285
...      Meltdown Build playing White: 56 - 25 - 62  [0.608] 143
...      Meltdown Build playing Black: 57 - 27 - 58  [0.606] 142
...      White vs Black: 83 - 82 - 120  [0.502] 285
Elo difference: 75.5 +/- 30.9, LOS: 100.0 %, DrawRatio: 42.1 %
SPRT: llr 2.96 (100.4%), lbound -2.94, ubound 2.94 - H1 was accepted

### DIFF
--- a/src/evaluation/move_ordering.h
+++ b/src/evaluation/move_ordering.h
@@ -57,15 +57,22 @@ public:
 
     constexpr void sortMoves(const BitBoard& board, movegen::ValidMoves& moves, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt)
     {
-        /* use stable sort to keep determinism */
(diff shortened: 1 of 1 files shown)


hansbinderup:
Gonna merge this so it's included in the nightly :)

PR #24, opened 2025-03-20
move_gen: Implement half-search for `getPieceAtSquare`

clarashepherd:
Colour is always known for these calls, so only search for pieces of that colour.

### DIFF
--- a/src/bit_board.h
+++ b/src/bit_board.h
@@ -47,15 +47,33 @@ struct BitBoard {
         occupation[Both] = occupation[White] | occupation[Black];
     }
 
-    constexpr std::optional<Piece> getPieceAtSquare(uint64_t square) const
+    template<Player player>
+    constexpr std::optional<Piece> getTargetAtSquare(uint64_t square) const
     {
-        for (const auto piece : magic_enum::enum_values<Piece>()) {
-            if (square & pieces[piece]) {
-                return piece;
+        if constexpr (player == PlayerWhite) {
+            for (const auto piece : s_blackPieces) {
+                if (square & pieces[piece]) {
+                    return piece;
+                }
+            }
+        } else {
+            for (const auto piece : s_whitePieces) {
+                if (square & pieces[piece]) {
+                    return piece;
+                }
             }
         }
 
         return std::nullopt;
+    }
+
+    // Helper: calling within loops will mean redundant colour checks
+    constexpr std::optional<Piece> getTargetAtSquare(uint64_t square, Player player) const
+    {
+        if (player == PlayerWhite)
+            return getTargetAtSquare<PlayerWhite>(square);
+        else
+            return getTargetAtSquare<PlayerBlack>(square);
     }
 
     constexpr bool isQuietPosition() const
@@ -79,4 +97,3 @@ struct BitBoard {
     uint32_t fullMoves {};
     uint32_t halfMoves {};
 };
-
--- a/src/engine/move_handling.h
+++ b/src/engine/move_handling.h
@@ -60,10 +60,11 @@ constexpr void performCastleMove(BitBoard& board, const movegen::Move& move, uin
     }
 }
 
+template<Player player>
 constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move, uint64_t& hash)
 {
-    const bool isWhite = board.player == PlayerWhite;
-    const auto type = isWhite ? WhitePawn : BlackPawn;
+    constexpr bool isWhite = player == PlayerWhite;
+    constexpr auto type = isWhite ? WhitePawn : BlackPawn;
 
     // first clear to be promoted pawn
     uint64_t& pawns = isWhite ? board.pieces[WhitePawn] : board.pieces[BlackPawn];
@@ -71,7 +72,7 @@ constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move, 
 
     /* clear piece that will be taken if capture */
     if (move.isCapture()) {
-        if (const auto victim = board.getPieceAtSquare(move.toSquare())) {
+        if (const auto victim = board.getTargetAtSquare<player>(move.toSquare())) {
             clearPiece(board.pieces[victim.value()], move.toPos(), victim.value(), hash);
         }
     }
@@ -80,19 +81,19 @@ constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move, 
     case PromotionNone:
         return;
     case PromotionQueen: {
-        const auto type = isWhite ? WhiteQueen : BlackQueen;
+        constexpr auto type = isWhite ? WhiteQueen : BlackQueen;
         setPiece(board.pieces[type], move.toPos(), type, hash);
     } break;
     case PromotionKnight: {
-        const auto type = isWhite ? WhiteKnight : BlackKnight;
+        constexpr auto type = isWhite ? WhiteKnight : BlackKnight;
         setPiece(board.pieces[type], move.toPos(), type, hash);
     } break;
     case PromotionBishop: {
-        const auto type = isWhite ? WhiteBishop : BlackBishop;
+        constexpr auto type = isWhite ? WhiteBishop : BlackBishop;
         setPiece(board.pieces[type], move.toPos(), type, hash);
     } break;
     case PromotionRook: {
-        const auto type = isWhite ? WhiteRook : BlackRook;
+        constexpr auto type = isWhite ? WhiteRook : BlackRook;
         setPiece(board.pieces[type], move.toPos(), type, hash);
     } break;
     }
@@ -177,6 +178,7 @@ constexpr static inline BoardPosition enpessantCapturePosition(BoardPosition pos
     }
 }
 
+template<Player player>
 constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move, uint64_t& hash)
 {
     BitBoard newBoard = board;
@@ -187,7 +189,7 @@ constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move,
     if (move.isCastleMove()) {
         performCastleMove(newBoard, move, hash);
     } else if (move.takeEnPessant()) {
-        if (newBoard.player == PlayerWhite) {
+        if constexpr (player == PlayerWhite) {
             movePiece(newBoard.pieces[WhitePawn], fromPos, toPos, WhitePawn, hash);
             clearPiece(newBoard.pieces[BlackPawn], enpessantCapturePosition(toPos, PlayerWhite), BlackPawn, hash);
         } else {
@@ -195,10 +197,10 @@ constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move,
             clearPiece(newBoard.pieces[WhitePawn], enpessantCapturePosition(toPos, PlayerBlack), WhitePawn, hash);
         }
     } else if (move.isPromotionMove()) {
-        performPromotionMove(newBoard, move, hash);
+        performPromotionMove<player>(newBoard, move, hash);
     } else {
         if (move.isCapture()) {
-            if (const auto victim = board.getPieceAtSquare(move.toSquare())) {
+            if (const auto victim = board.getTargetAtSquare<player>(move.toSquare())) {
                 clearPiece(newBoard.pieces[victim.value()], move.toPos(), victim.value(), hash);
             }
         }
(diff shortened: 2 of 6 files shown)


src/engine/move_handling.h
```diff
@@ -71,7 +71,7 @@ constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move,
 
     /* clear piece that will be taken if capture */
     if (move.isCapture()) {
-        if (const auto victim = board.getPieceAtSquare(move.toSquare())) {
+        if (const auto victim = (board.player == PlayerWhite) ? board.getPieceAtSquare<PlayerWhite>(move.toSquare()) : board.getPieceAtSquare<PlayerBlack>(move.toSquare())) {
```
clarashepherd, line 74: In the current implementation, we don't save on any checks for the colour, because the function calling `getPieceAtSquare` has not already checked for the colour.

It seems that if we want gains from the templating, we'd have to pass the pattern higher.

What do you think?

src/engine/move_handling.h
```diff
@@ -202,7 +202,7 @@ constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move,
         performPromotionMove(newBoard, move, hash);
     } else {
         if (move.isCapture()) {
-            if (const auto victim = board.getPieceAtSquare(move.toSquare())) {
+            if (const auto victim = (newBoard.player == PlayerWhite) ? board.getPieceAtSquare<PlayerWhite>(move.toSquare()) : board.getPieceAtSquare<PlayerBlack>(move.toSquare())) {
```
clarashepherd, line 205: This is the only exception to my above comment, because there's one other check for the colour in this function.

However I tried rewriting so the check was the outer of the nested conditionals, and it ended up with quite a few lines of duplication that IMO isn't worth the possible benefits at the current time.

src/evaluation/history_moves.h
```diff
@@ -21,7 +21,7 @@ class HistoryMoves {
             return; // nothing to do
         }
 
-        const auto movePiece = board.getPieceAtSquare(move.fromSquare());
```
hansbinderup, line 24: This is nice - good catch :+1:

hansbinderup (commented):
A few comments, otherwise looks good :100:

src/bit_board.h
```diff
@@ -47,11 +47,20 @@ struct BitBoard {
         occupation[Both] = occupation[White] | occupation[Black];
     }
 
+    template<Player player>
     constexpr std::optional<Piece> getPieceAtSquare(uint64_t square) const
```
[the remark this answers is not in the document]
clarashepherd: Followed this pattern and added helpers to the new templated functions, wherever a helper was...helpful.

src/bit_board.h
```diff
@@ -47,11 +47,20 @@ struct BitBoard {
...
+            for (const auto piece : s_blackPieces) {
+                if (square & pieces[piece]) {
+                    return piece;
+                }
+            }
+        } else {
+            for (const auto piece : s_whitePieces) {
+                if (square & pieces[piece]) {
+                    return piece;
+                }
```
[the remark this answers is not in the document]
clarashepherd: Done.

src/engine/move_handling.h
clarashepherd: I agree and added templating there, but also saw an opportunity for more significant optimization:

Profiling showed that the heaviest use of `getPieceAtSquare` was in `sortMoves`, where the latter is called for each valid move.

By templating `moveScore`, I've extracted the colour check from the 'for each move' loop, which should get us closer to a genuine half search. It only added a few lines to `sortMoves` so even if the instruction cache is a little bit more full, the gains should be a a net positive.

What do you think?

src/evaluation/history_moves.h
```diff
@@ -21,7 +21,7 @@ class HistoryMoves {
...
+        const auto movePiece = (board.player == PlayerWhite) ? board.getPieceAtSquare<PlayerWhite>(move.fromSquare()) : board.getPieceAtSquare<PlayerBlack>(move.fromSquare());
```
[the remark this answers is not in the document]
clarashepherd: Thanks, I'll follow this going forwards :)

src/evaluation/move_ordering.h
```diff
@@ -105,14 +105,18 @@ class MoveOrdering {
...
-            const auto attacker = board.getPieceAtSquare(move.fromSquare());
             const auto killerMoves = m_killerMoves.get(ply);
             if (move == killerMoves.first)
                 return ScoreKillerMove;
             else if (move == killerMoves.second)
                 return ScoreKillerMove - 1000;
-            else if (attacker.has_value()) {
-                return ScoreHistoryMove + m_historyMoves.get(attacker.value(), move.toPos());
+            else {
+                std::optional<Piece> attacker = (board.player == PlayerWhite)
```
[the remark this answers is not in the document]
clarashepherd: Nice, I get it :) I'll follow this going forwards.

src/evaluation/see_swap.h
```diff
@@ -81,11 +81,12 @@ class SeeSwap {
         /* piece that will track the scoring of next piece */
         Piece nextPiece = move.piece();
 
-        if (const auto initialPiece = board.getPieceAtSquare(move.toSquare())) {
+        Player player = board.player;
+
+        if (const auto initialPiece = (player == PlayerWhite) ? board.getPieceAtSquare<PlayerWhite>(move.toSquare()) : board.getPieceAtSquare<PlayerBlack>(move.toSquare())) {
```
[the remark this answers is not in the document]
clarashepherd: Now redundant on these specific lines.

src/engine/move_handling.h
```diff
@@ -246,6 +248,15 @@ constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move,
...
 }
 
+// Helper: using inside loops means redundant colour checks
+constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move, uint64_t& hash, Player player)
+{
+    if (player == PlayerWhite)
+        return performMove<PlayerWhite>(board, move, hash);
+    else
+        return performMove<PlayerBlack>(board, move, hash);
+}
```
hansbinderup, line 258: This could maybe be a cool way to implement it without adding redundancy? It will have same signature, but without the template, so you can branch at compile time using the template or runtime when not :)

Maybe it's less explicit? 

```suggestion
constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move, uint64_t& hash)
{
    if (board.player == PlayerWhite)
        return performMove<PlayerWhite>(board, move, hash);
    else
        return performMove<PlayerBlack>(board, move, hash);
}
```

src/engine/move_handling.h
```diff
@@ -60,6 +60,7 @@ constexpr void performCastleMove(BitBoard& board, const movegen::Move& move, uin
     }
 }
 
+template<Player player>
 constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move, uint64_t& hash)
 {
     const bool isWhite = board.player == PlayerWhite;
```
hansbinderup, line 66: These can be evaluated at compile time as well now :fire: 
```suggestion
    constexpr bool isWhite = player == PlayerWhite;
    constexpr auto type = isWhite ? WhitePawn : BlackPawn;
    uint64_t& pawns = board.pieces[type];
```

src/engine/move_handling.h
```diff
@@ -232,7 +234,7 @@ constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move,
     newBoard.attacks[PlayerBlack] = attackgen::getAllAttacks<PlayerBlack>(newBoard);
 
     /* player making the move is black -> inc full moves */
-    if (board.player == PlayerBlack)
+    if (player == PlayerBlack)
```
hansbinderup, line 237:
```suggestion
    if constexpr (player == PlayerBlack)
```

src/evaluation/move_ordering.h
```diff
@@ -105,25 +117,36 @@ class MoveOrdering {
...
 
+    // Helper: calling inside loops will mean redundant colour checks
+    constexpr int32_t moveScore(Player player, const BitBoard& board, const movegen::Move& move, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt) const
+    {
+        if (player == PlayerWhite) {
+            return moveScore<PlayerWhite>(board, move, ply, ttMove);
+        } else {
+            return moveScore<PlayerBlack>(board, move, ply, ttMove);
+        }
+    }
```
hansbinderup, line 144: Same applies here:
```suggestion
    // Helper: calling inside loops will mean redundant colour checks
    constexpr int32_t moveScore(const BitBoard& board, const movegen::Move& move, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt) const
    {
        if (board.player == PlayerWhite) {
            return moveScore<PlayerWhite>(board, move, ply, ttMove);
        } else {
            return moveScore<PlayerBlack>(board, move, ply, ttMove);
        }
    }
```

src/uci_handler.h
```diff
@@ -107,7 +107,7 @@ class UciHandler {
                 const auto move = parsing::moveFromString(s_board, moveSv.value_or(sv));
 
                 if (move.has_value()) {
-                    s_board = engine::performMove(s_board, move.value(), hash);
+                    s_board = s_board.player == PlayerWhite ? engine::performMove<PlayerWhite>(s_board, move.value(), hash) : engine::performMove<PlayerBlack>(s_board, move.value(), hash);
```
hansbinderup, line 110: Helper can be used here
```suggestion
                    s_board = engine::performMove(s_board, move.value(), hash);
```

src/evaluation/move_ordering.h
```diff
@@ -60,19 +60,31 @@ class MoveOrdering {
...
-            while (j > 0 && moveScore(board, moves[j - 1], ply, ttMove) < keyScore) {
-                moves[j] = moves[j - 1];
-                --j;
+
+            if (board.player == PlayerWhite) {
+                int keyScore = moveScore<PlayerWhite>(board, key, ply, ttMove);
+                size_t j = i;
+                while (j > 0 && moveScore<PlayerWhite>(board, moves[j - 1], ply, ttMove) < keyScore) {
+                    moves[j] = moves[j - 1];
+                    --j;
```
hansbinderup, line 69: Maybe this could be added to a lambda? Not sure if it will add runtime overhead - might need to godbolt to see if compiler can optimize it. Otherwise, fine with me :+1:

hansbinderup (commented):
Really nice with the further optimizations! I found a few more that you might as well include :+1:

src/evaluation/move_ordering.h
clarashepherd: Now that you mention it, I think the neatest solution here is another template and helper. It's also more optimised than the above lines, because the conditional for colour is entirely out of the loops over moves.

src/engine/move_handling.h
```diff
@@ -60,18 +60,19 @@ constexpr void performCastleMove(BitBoard& board, const movegen::Move& move, uin
...
+template<Player player>
 constexpr void performPromotionMove(BitBoard& board, const movegen::Move& move, uint64_t& hash)
 {
-    const bool isWhite = board.player == PlayerWhite;
-    const auto type = isWhite ? WhitePawn : BlackPawn;
+    constexpr bool isWhite = player == PlayerWhite;
+    constexpr auto type = isWhite ? WhitePawn : BlackPawn;
 
     // first clear to be promoted pawn
     uint64_t& pawns = isWhite ? board.pieces[WhitePawn] : board.pieces[BlackPawn];
```
hansbinderup, line 70: No need for the tenary. You can just use type.
You can also constexpr the types below. Can't comment/suggest code from the phone :/

src/engine/move_handling.h
```diff
@@ -181,6 +182,7 @@ constexpr static inline BoardPosition enpessantCapturePosition(BoardPosition pos
     }
 }
 
+template<Player player>
 constexpr BitBoard performMove(const BitBoard& board, const movegen::Move& move, uint64_t& hash)
```
hansbinderup, line 186: There is more 'board.player' we can constexpr in this method :)

src/evaluation/move_ordering.h
```diff
@@ -105,25 +117,35 @@ class MoveOrdering {
...
-            const auto attacker = board.getPieceAtSquare(move.fromSquare());
             const auto killerMoves = m_killerMoves.get(ply);
             if (move == killerMoves.first)
                 return ScoreKillerMove;
             else if (move == killerMoves.second)
                 return ScoreKillerMove - 1000;
-            else if (attacker.has_value()) {
-                return ScoreHistoryMove + m_historyMoves.get(attacker.value(), move.toPos());
+            else {
+                auto attacker = board.getTargetAtSquare<player>(move.fromSquare());
```
hansbinderup, line 126: Nitpick: can be const

hansbinderup (approved):
This is starting to add so much! Really good! Just found a few more constexpr optimizations, if you bother to include it :)

PR #29, opened 2025-03-22
Implement atomic TT hashing

clarashepherd:


src/engine/tt_hash_table.h
```diff
@@ -88,34 +100,39 @@ class TtHashTable {
...
+        newEntry.score = score;
+        newEntry.flag = flag;
+        newEntry.depth = depth;
+        newEntry.generation = s_currentGeneration;
+        newEntry.move = move;
+
+        auto& entryWrapper = s_ttHashTable[key % s_ttHashSize];
+        auto oldEntry = entryWrapper.load(std::memory_order_acquire);
+
+        while (!entryWrapper.compare_exchange_weak(oldEntry, newEntry, std::memory_order_release, std::memory_order_acquire)) {
```
clarashepherd, line 120: Add `is_lock_free` compile-time check

src/engine/tt_hash_table.h
```diff
@@ -14,7 +15,7 @@ enum TtHashFlag : uint8_t {
     TtHashBeta = 2,
 };
 
-struct TtHashEntry {
+struct alignas(64) TtHashEntry {
```
clarashepherd, line 18: Recheck alignment: could compact further. Look into cache lines.

src/engine/tt_hash_table.h
clarashepherd, line 120: Look up memory order: can we use relaxed and guarantee atomicity?

src/engine/tt_hash_table.h
```diff
@@ -88,34 +100,39 @@ class TtHashTable {
...
-        entry.move = move;
     }
 
 private:
     constexpr static inline std::size_t s_hashTableSizeMb { 128 };
     constexpr static inline std::size_t s_ttHashSize { (s_hashTableSizeMb * 1024 * 1024) / sizeof(TtHashEntry) };
     constexpr static inline uint8_t s_generationMask { 0xF }; /* 4-bit wrapping generation (max 16 generations allowed) */
 
-    static inline std::array<TtHashEntry, s_ttHashSize> s_ttHashTable; /* TODO: make dynamic so size can be adjusted from UCI */
+    static inline std::array<std::atomic<TtHashEntry>, s_ttHashSize> s_ttHashTable; /* TODO: make dynamic so size can be adjusted from UCI */
```
clarashepherd, line 134: Exercise: try removing atomic read/write, to just change `entry` directly as before. Hash can't be corrupted. Might go down false-good branches but will be corrected. Compare overheads.

src/engine/tt_hash_table.h
```diff
@@ -46,40 +47,51 @@ class TtHashTable {
...
+        if (oldEntry.generation != s_currentGeneration) {
+            TtHashEntry newEntry = oldEntry;
+            newEntry.key = 0;
             if (s_hashCount > 0) {
+                // TODO make atomic
+                // Not sure how?
                 --s_hashCount;
             }
+            // Try to invalidate the entry. May not invalidate entry
+            if (!entryWrapper.compare_exchange_weak(oldEntry, newEntry, std::memory_order_release, std::memory_order_acquire)) {
```
clarashepherd, line 68: Redundant. Just an atomic load.

src/engine/tt_hash_table.h
```diff
@@ -46,40 +47,51 @@ class TtHashTable {
...
+        auto oldEntry = entryWrapper.load(std::memory_order_acquire);
 
-        if (entry.key != key) {
+        if (oldEntry.key != key) {
             return std::nullopt;
         }
 
         /* Entry is outdated (from a previous generation) */
-        if (entry.generation != s_currentGeneration) {
-            entry.key = 0; // Invalidate the entry
```
clarashepherd, line 57: No need to invalidate, just ignore.

src/engine/tt_hash_table.h
```diff
@@ -46,40 +47,51 @@ class TtHashTable {
     using ProbeResult = std::pair<int32_t, movegen::Move>;
     constexpr static std::optional<ProbeResult> probe(uint64_t key, int32_t alpha, int32_t beta, uint8_t depth, uint8_t ply)
     {
-        auto& entry = s_ttHashTable[key % s_ttHashSize];
+        auto& entryWrapper = s_ttHashTable[key % s_ttHashSize];
+        // Load all at once atomically
+        auto oldEntry = entryWrapper.load(std::memory_order_acquire);
```
clarashepherd, line 52: Rename to `loadedEntry`. Pattern makes sense though: load once so not updating during control flow.

src/engine/tt_hash_table.h
```diff
@@ -46,40 +47,51 @@ class TtHashTable {
...
         /* Entry is outdated (from a previous generation) */
-        if (entry.generation != s_currentGeneration) {
-            entry.key = 0; // Invalidate the entry
+        if (oldEntry.generation != s_currentGeneration) {
+            TtHashEntry newEntry = oldEntry;
+            newEntry.key = 0;
             if (s_hashCount > 0) {
+                // TODO make atomic
+                // Not sure how?
                 --s_hashCount;
```
clarashepherd, line 65: Nice-to-have value for GUI only. Currently modifying within reader. Could be racy but not worth the overhead for a ballpark value.

Might lead to overflow, let's see if it does.

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+    constexpr static inline uint64_t s_scoreMask { 0b11111111111111111 };
+    constexpr static inline uint64_t s_scoreShift { 10 };
+
+    constexpr static inline uint64_t s_generationMask { 0b11111111 };
+    constexpr static inline uint64_t s_generationShift { 27 };
+
+    constexpr static inline uint64_t s_moveMask { 0b11111111111111111111111111 };
+    constexpr static inline uint64_t s_moveShift { 35 };
+
+    uint64_t m_data;
```
hansbinderup, line 82: For default construction
```suggestion
    uint64_t m_data {};
```

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+        return (m_data >> s_generationShift) & s_generationMask;
+    }
+
+    constexpr inline uint32_t getMoveData() const
+    {
+        return (m_data >> s_moveShift) & s_moveMask;
+    }
+
+private:
+    constexpr static inline uint64_t s_depthMask { 0b1111111 };
```
hansbinderup, line 68: Is 8 bit value not 7
Currently it's just limited to 128 (most engines use 255)

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+
+    constexpr static inline uint64_t s_flagMask { 0b111 };
+    constexpr static inline uint64_t s_flagShift { 7 };
+
+    constexpr static inline uint64_t s_scoreMask { 0b11111111111111111 };
+    constexpr static inline uint64_t s_scoreShift { 10 };
+
+    constexpr static inline uint64_t s_generationMask { 0b11111111 };
```
hansbinderup, line 76: We only support up to 16 generation so 4 bits should be enough, right?

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+    constexpr static inline uint64_t s_generationShift { 27 };
+
+    constexpr static inline uint64_t s_moveMask { 0b11111111111111111111111111 };
+    constexpr static inline uint64_t s_moveShift { 35 };
```
hansbinderup, line 80: This will overflow, hence the warning

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+    constexpr static inline uint64_t s_flagMask { 0b111 };
+    constexpr static inline uint64_t s_flagShift { 7 };
+
+    constexpr static inline uint64_t s_scoreMask { 0b11111111111111111 };
+    constexpr static inline uint64_t s_scoreShift { 10 };
+
+    constexpr static inline uint64_t s_generationMask { 0b11111111 };
+    constexpr static inline uint64_t s_generationShift { 27 };
+
+    constexpr static inline uint64_t s_moveMask { 0b11111111111111111111111111 };
```
hansbinderup, line 79: This is very verbose 😀 
Maybe just use the hex values for masks (in general)

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+ * score: 17 bits (see board_defs.h)
+ * generation: 8 bits
+ * move: 26 bits (see move_types.h)
+ * */
+
+class TtHashEntryData {
+public:
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshift-count-overflow"
+    explicit TtHashEntryData(uint64_t depth = 0, uint64_t flag = TtHashExact, int64_t score = 0, uint64_t generation = 0, uint64_t move = {})
```
hansbinderup, line 31: No need for default args when data itself is default initialized. Probably the issue you tried to solve here?:)
```suggestion
    explicit TtHashEntryData(uint8_t depth = 0, uint8_t flag = TtHashExact, int32_t score = 0, uint8_t generation = 0, uint32_t move = {})
```

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,87 @@ enum TtHashFlag : uint8_t {
...
+ * flag: 3 bits, i.e. 0000 ... 0000 0011 1000 0000
+ * score: 17 bits (see board_defs.h)
+ * generation: 8 bits
+ * move: 26 bits (see move_types.h)
+ * */
+
+class TtHashEntryData {
+public:
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshift-count-overflow"
```
hansbinderup, line 30: With my comments you should probably on longer see this warning

src/engine/tt_hash_table.h
hansbinderup, line 31: Pretty sure that this doesn't work with the signed value (score). We need to store the signedness somewhere - hence the overflow warning.

There's a few ways to do it, eg:

1. Have a signed bit in our encoded data (1 if score is negative and then store score as abs value).

2. Offset the absolute score and do the opposite when decoding.

hansbinderup (commented):
Some initial comments before we go climbing :)

Good news: I'm pretty sure that it's not actually working
Bad news: Well, same 😆

src/engine/tt_hash_table.h
clarashepherd: Does this somewhat contradict your previous comment? Although I agree that we have more space in the generation than the depth :)

hansbinderup: Depth and generation are not related :)

src/engine/tt_hash_table.h
hansbinderup: Will the warning go away if we mask the value?

src/engine/tt_hash_table.h
clarashepherd: Good spot, thanks.

src/engine/tt_hash_table.h
clarashepherd: OK. I'll look into this outside of the PR.

src/engine/tt_hash_table.h
clarashepherd: It's gone away anyway with the more optimal compression of the generation.

src/engine/tt_hash_table.h
clarashepherd: I agree, done!

src/engine/tt_hash_table.h
clarashepherd: Nice spot, changed.

src/engine/tt_hash_table.h
clarashepherd: That's right. I've removed the ignore.

src/engine/tt_hash_table.h
```diff
@@ -14,20 +15,95 @@ enum TtHashFlag : uint8_t {
...
+              depth
+              | flag << s_flagShift
+              // Most significant bit: sign (1 if negative)
+              | (score > 0 ? score : -score + (1 << 16)) << s_scoreShift
+              | generation << s_generationShift
+              | move << s_moveShift)
+    {
+    }
+
+    constexpr inline uint8_t getDepth() const
```
hansbinderup, line 42: constexpr is default inline. I added inline other places because I was not aware that it was default, so my bad :)

```suggestion
    constexpr uint8_t getDepth() const
```

hansbinderup (approved):
A few minors and otherwise good to go :100:

src/engine/tt_hash_table.h
```diff
@@ -88,24 +167,25 @@ class TtHashTable {
...
         if (score < -s_mateScore)
             score -= ply;
         if (score > s_mateScore)
             score += ply;
 
-        if (entry.key == 0) {
+        auto& entry = s_ttHashTable[key % s_ttHashSize];
+
+        if (entry.key.load(std::memory_order_relaxed) == 0) {
             ++s_hashCount;
```
hansbinderup, line 179: We should probably make this one atomic as well

clarashepherd:
Tournament of 500 games, st=0.1. Elo difference: 11.8 +/- 20.4.

PR #31, opened 2025-03-23
static_evaluation: add extended tapered evaluation + performance improvements

hansbinderup:
Our current static evaluation (renamed now) does not handle game phases very well. 
We should now be able to control game phase scores in a more dynamic way.
Also the performance from only looping each piece once is quite massive:

```
Score of Meltdown Build vs Meltdown Baseline: 275 - 200 - 281  [0.550] 756
...      Meltdown Build playing White: 156 - 92 - 131  [0.584] 379
...      Meltdown Build playing Black: 119 - 108 - 150  [0.515] 377
...      White vs Black: 264 - 211 - 281  [0.535] 756
Elo difference: 34.6 +/- 19.7, LOS: 100.0 %, DrawRatio: 37.2 %
SPRT: llr 2.96 (100.4%), lbound -2.94, ubound 2.94 - H1 was accepted
```

### DIFF
--- a/src/evaluation/game_phase.h
+++ b/src/evaluation/game_phase.h
@@ -1,0 +1,71 @@
+#pragma once
+
+#include "board_defs.h"
+#include "magic_enum/magic_enum.hpp"
+
+#include <algorithm>
+#include <cstdint>
+
+namespace evaluation::gamephase {
+
+namespace {
+
+constexpr uint8_t s_mgLimit { 24 };
+
+}
+
+constexpr std::array<uint8_t, magic_enum::enum_count<Piece>()> s_materialPhaseScore = {
+    0, 1, 1, 2, 4, 0, /* white pieces */
+    0, 1, 1, 2, 4, 0 /* black pieces */
+};
+
+enum Phases : uint8_t {
+    GamePhaseMg = 0,
+    GamePhaseEg,
+};
+
+struct Score {
+    int32_t mg {};
+    int32_t eg {};
+    uint8_t materialScore {};
+
+    constexpr Score operator+(const Score& other) const noexcept
+    {
+        return { static_cast<int32_t>(mg + other.mg), static_cast<int32_t>(eg + other.eg), static_cast<uint8_t>(materialScore + other.materialScore) };
+    }
+
+    constexpr Score& operator+=(const Score& other) noexcept
+    {
+        mg += other.mg;
+        eg += other.eg;
+        materialScore += other.materialScore;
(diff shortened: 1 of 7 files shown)


hansbinderup:
@clarashepherd I couldn't reopen the old PR as I force pushed after closing it, sorry!
I'll respond to your comments here instead :)

src/evaluation/game_phase.h
```diff
@@ -0,0 +1,82 @@
...
+    uint8_t eg;
+};
+
+[[nodiscard]] constexpr Weight computeWeight(uint8_t materialPhase)
+{
+    /* we have to limit the score in case of early promotions */
+    const uint8_t mgPhase = std::min(materialPhase, s_mgLimit);
+    const uint8_t egPhase = s_mgLimit - mgPhase;
+
+    return { mgPhase, egPhase };
```
hansbinderup, line 26: @clarashepherd I see your point, but I kinda prefer isolating the weight handling here. I think that's more readable? Or maybe I'm missing something? Please throw some physics at me! :)

clarashepherd: I'm not sure we're actually isolating the weight handling here though - we're just producing two numbers that aren't meaningful until they're used in a weighting formula.

This isn't a big point, and I'm also happy with the two weights here. Use your judgement, this isn't a blocker for me :)

src/evaluation/positioning.h
```diff
@@ -5,136 +5,210 @@
...
 namespace evaluation {
 
 namespace {
 
-constexpr int32_t s_doublePawnPenalty { -10 };
-constexpr int32_t s_isolatedPawnPenalty { -10 };
-constexpr auto s_passedPawnBonus = std::to_array<int32_t>({ 0, 10, 30, 50, 75, 100, 150, 200 });
-constexpr int32_t s_semiOpenFileScore { 10 };
-constexpr int32_t s_openFileScore { 15 };
+/* TODO: find a way to tune these parameters in an automated way - currently they're just guesses
```
clarashepherd, line 21: Does this point to the possibility of tuning more generally throughout the engine? AFAI can see the only parameters we're manually obtaining are the magic numbers for sliding pieces.

hansbinderup: Yes exactly! We need to build a tuner that can evaluate positions, based on a known dataset, and then optimize the parameters based on that. Something else todo 🙃

clarashepherd:
Very minor comments, otherwise good to go :zap:

PR #36, opened 2025-03-28
Add futility and razoring pruning

hansbinderup:
See each commit for details.

A while ago since we added pruning. About time to collect some easy elo :))

Results speak for themselves:
```
Score of Meltdown Build vs Meltdown Baseline: 130 - 63 - 115  [0.609] 308
Finished game 306 (Meltdown Baseline vs Meltdown Build): * {No result}
Score of Meltdown Build vs Meltdown Baseline: 130 - 63 - 115  [0.609] 308
...      Meltdown Build playing White: 73 - 30 - 52  [0.639] 155
...      Meltdown Build playing Black: 57 - 33 - 63  [0.578] 153
...      White vs Black: 106 - 87 - 115  [0.531] 308
Elo difference: 76.8 +/- 31.0, LOS: 100.0 %, DrawRatio: 37.3 %
SPRT: llr 2.98 (101.1%), lbound -2.94, ubound 2.94 - H1 was accepted
```

### DIFF
--- a/src/evaluation/evaluator.h
+++ b/src/evaluation/evaluator.h
@@ -362,6 +362,24 @@ private:
             }
         }
 
+        /* https://www.chessprogramming.org/Razoring (Strelka) */
+        if (!isPv && !isChecked && depth <= s_reductionLimit) {
+            int32_t score = staticEval + s_razorMarginShallow;
+            if (score < beta) {
+                if (depth == 1) {
+                    int32_t newScore = quiesence(board, alpha, beta);
+                    return (newScore > score) ? newScore : score;
+                }
+
+                score += s_razorMarginDeep;
+                if (score < beta && depth <= s_razorDeepReductionLimit) {
+                    const int32_t newScore = quiesence(board, alpha, beta);
+                    if (newScore < beta)
+                        return (newScore > score) ? newScore : score;
+                }
+            }
+        }
+
         bool tbMoves = false;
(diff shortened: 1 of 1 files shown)


clarashepherd (approved):
Awesome, let's get this in :1st_place_medal:

sroelants:
A word of warning about testing multiple changes at once: it's really hard to tell from this whether both of these changes gained elo individually. Even if the implementation is correct, the exact values you use for the cutoff margins are so sensitive to the specific engine. It's totally possible, for example, that RFC gained you 100+ elo, but razoring actually lost you 20 elo because the margins were too aggressive, and you'd have no way of knowing from this result alone.

In theory, it's best to always test changes individually. In practice, that sucks, because it means running _a lot_ of games. :(

hansbinderup:
> A word of warning about testing multiple changes at once: it's really hard to tell from this whether both of these changes gained elo individually. Even if the implementation is correct, the exact values you use for the cutoff margins are so sensitive to the specific engine. It's totally possible, for example, that RFC gained you 100+ elo, but razoring actually lost you 20 elo because the margins were too aggressive, and you'd have no way of knowing from this result alone.
> 
> 
> 
> In theory, it's best to always test changes individually. In practice, that sucks, because it means running _a lot_ of games. :(

Thanks for the heads up! I'll keep that in mind and run some more tests.. Test test test - better get used to it 🤓

PR #38, opened 2025-03-30
WIP: implement lazy SMP

clarashepherd:


src/thread_pool/thread_pool.h
```diff
@@ -0,0 +1,116 @@
+#include <condition_variable>
+#include <functional>
+#include <iostream>
+#include <mutex>
+#include <queue>
+#include <thread>
+
+namespace thread_pool {
```
clarashepherd, line 8: Made a generic an inefficient thread pool for now, while we're prototyping.

PR #45, opened 2025-04-06
Clara/implement lazy smp

clarashepherd:
See commit messages for details :)

### DIFF
--- a/src/evaluation/evaluator.h
+++ b/src/evaluation/evaluator.h
@@ -2,6 +2,7 @@
 
 #include "engine/thread_pool.h"
 #include "engine/tt_hash_table.h"
+#include "evaluation/move_vote_map.h"
 #include "evaluation/searcher.h"
 #include "time_manager.h"
 
@@ -16,9 +17,34 @@ namespace evaluation {
 
 class Evaluator {
 public:
+    Evaluator()
+    {
+        resizeSearchers(1);
+    };
+
+    void resizeSearchers(uint8_t size)
+    {
+        if (size == m_searchers.size()) {
+            return;
+        }
+
+        else {
+            for (uint8_t i = m_searchers.size(); i < size; i++) {
+                m_searchers.emplace_back(std::make_unique<Searcher>());
+            }
+        }
+
+        m_searchers.resize(size);
+        m_threadPool.resize(size + 2);
+    }
+
     constexpr uint64_t getNodes() const
     {
-        return m_searcher.getNodes();
+        uint64_t totalNodes {};
+        for (auto& searcher : m_searchers) {
+            totalNodes += searcher->getNodes();
+        }
+        return totalNodes;
     }
 
     constexpr movegen::Move getBestMove(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
@@ -33,7 +59,9 @@ public:
 
         const uint64_t hash = engine::generateHashKey(board);
 
-        m_searcher.setHashKey(hash);
+        for (auto& searcher : m_searchers) {
+            searcher->setHashKey(hash);
+        }
 
         return scanForBestMove(depthInput.value_or(s_maxSearchDepth), board);
     }
@@ -57,9 +85,8 @@ public:
 
             fmt::print("bestmove {}", move);
             if (m_ponderingEnabled) {
-                const auto ponderMove = m_searcher.getPonderMove();
-                if (ponderMove.has_value()) {
-                    fmt::print(" ponder {}", ponderMove.value());
+                if (m_ponderMove.has_value()) {
+                    fmt::print(" ponder {}", m_ponderMove.value());
                 }
             }
 
@@ -97,7 +124,9 @@ public:
 
     constexpr void printEvaluation(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
     {
-        m_searcher.printEvaluation(board, depthInput);
+        for (const auto& searcher : m_searchers) {
+            searcher->printEvaluation(board, depthInput);
+        }
     }
 
     /*
@@ -107,18 +136,25 @@ public:
     void resetTiming()
     {
         TimeManager::reset();
-        m_searcher.resetNodes();
+
+        for (auto& searcher : m_searchers) {
+            searcher->resetNodes();
+        }
     }
 
     void reset()
     {
         resetTiming(); // full reset required
-        m_searcher.reset();
+        for (auto& searcher : m_searchers) {
+            searcher->reset();
+        }
     }
 
     void updateRepetition(uint64_t hash)
     {
-        m_searcher.updateRepetition(hash);
+        for (auto& searcher : m_searchers) {
+            searcher->updateRepetition(hash);
+        }
     }
 
     constexpr void printScoreInfo(Searcher* searcher, const BitBoard& board, int32_t score, uint8_t currentDepth)
@@ -143,6 +179,8 @@ public:
         fflush(stdout);
     }
 
+    constexpr static inline uint8_t s_maxThreads { 128 };
+
 private:
     constexpr movegen::Move scanForBestMove(uint8_t depth, const BitBoard& board)
     {
@@ -153,6 +191,7 @@ private:
          * iterative deeping - with aspiration window
          * https://web.archive.org/web/20070705134903/www.seanet.com/%7Ebrucemo/topics/aspiration.htm
          */
+        movegen::Move bestMove;
         uint8_t d = 1;
 
         while (d <= depth) {
@@ -160,21 +199,100 @@ private:
                 break;
             }
 
-            Searcher::setSearchStopped(false);
-            const int32_t score = m_searcher.startSearch(d, board, alpha, beta);
+            if (m_searchers.size() == 1) {
+                const auto& singleSearcher = m_searchers.front();
 
-            if ((score <= alpha) || (score >= beta)) {
-                alpha = s_minScore;
-                beta = s_maxScore;
+                Searcher::setSearchStopped(false);
+                const auto score = singleSearcher->startSearch(d, board, alpha, beta);
 
-                continue;
+                if ((score <= alpha) || (score >= beta)) {
+                    alpha = s_minScore;
+                    beta = s_maxScore;
+
+                    continue;
+                }
+
+                /* prepare window for next iteration */
+                alpha = score - s_aspirationWindow;
+                beta = score + s_aspirationWindow;
+
+                printScoreInfo(singleSearcher.get(), board, score, d);
+
+                bestMove = singleSearcher->getPvMove();
+                m_ponderMove = singleSearcher->getPonderMove();
+            } else {
+                Searcher::setSearchStopped(false);
+
+                /* Thread voting: https://www.chessprogramming.org/Lazy_SMP */
+                std::array<SearcherResult, s_maxThreads> searchResults {};
+                uint8_t numSearchResults {};
+
+                for (auto& searcher : m_searchers) {
+                    searcher->startSearchAsync(m_threadPool, d, board, alpha, beta);
+                }
+
+                for (auto& searcher : m_searchers) {
+                    const auto& result = searcher->getSearchResult();
+                    if (result.has_value()) {
+                        // If didn't fall out of window and wasn't an immediate early termination, push back to results
+                        if ((result.value().score > alpha) && (result.value().score < beta) && result.value().searchedDepth > 0) {
+                            searchResults.at(numSearchResults) = result.value();
+                            ++numSearchResults;
+                        }
+                    }
+                }
+
+                if (numSearchResults == 0) {
+                    alpha = s_minScore;
+                    beta = s_maxScore;
+
+                    continue;
+                }
+
+                m_movesVotes.clear();
+
+                for (uint8_t i = 0; i < numSearchResults; i++) {
+
+                    const auto& result = searchResults.at(i);
+
+                    const int32_t score = result.score;
+                    const uint8_t depth = result.searchedDepth;
+                    const auto move = result.pvMove;
+
+                    /* Can be tweaked for optimization */
+                    int64_t voteWeight = (score - s_minScore) * depth;
+
+                    m_movesVotes.insertOrIncrement(move, voteWeight);
+                }
+
+                int64_t maxVote = -std::numeric_limits<int64_t>::max();
+
+                for (const auto& [move, vote] : m_movesVotes) {
+                    if (vote > maxVote) {
+                        maxVote = vote;
+                        bestMove = move;
+                    }
+                }
+
+                uint8_t bestWinningDepth = 0;
+                SearcherResult bestWinningResult;
+
+                for (uint8_t i = 0; i < numSearchResults; i++) {
+                    if (searchResults.at(i).pvMove == bestMove) {
+                        if (depth > bestWinningDepth) {
+                            bestWinningResult = searchResults.at(i);
+                            bestWinningDepth = depth;
+                        }
+                    }
+                }
+
+                /* prepare window for next iteration */
+                alpha = bestWinningResult.score - s_aspirationWindow;
+                beta = bestWinningResult.score + s_aspirationWindow;
+
+                printScoreInfo(bestWinningResult.searcher, board, bestWinningResult.score, d);
+                m_ponderMove = bestWinningResult.searcher->getPonderMove();
             }
-
-            /* prepare window for next iteration */
-            alpha = score - s_aspirationWindow;
-            beta = score + s_aspirationWindow;
-
-            printScoreInfo(&m_searcher, board, score, d);
 
             /* only increment depth, if we didn't fall out of the window */
             d++;
@@ -183,15 +301,20 @@ private:
         /* in case we're scanning to a certain depth we need to ensure that we're stopping the time handler */
         stop();
 
-        return m_searcher.getPvMove();
+        return bestMove;
     }
 
     Searcher m_searcher {};
     std::atomic_bool m_killed { false };
 
-    ThreadPool m_threadPool { 1 }; /* single searcher */
+    ThreadPool m_threadPool { 3 }; /* Iterative deepening, time handler, default=1 searcher */
+
+    std::vector<std::unique_ptr<Searcher>> m_searchers {};
+    MoveVoteMap<s_maxThreads> m_movesVotes {};
+
     bool m_isPondering { false };
     bool m_ponderingEnabled { false };
+    std::optional<movegen::Move> m_ponderMove {};
 
     constexpr static inline uint8_t s_aspirationWindow { 50 };
 };
--- a/src/evaluation/move_vote_map.h
+++ b/src/evaluation/move_vote_map.h
@@ -1,0 +1,56 @@
+#pragma once
+
+#include "movegen/move_types.h"
+
+namespace evaluation {
+
+template<size_t maxSize>
+class MoveVoteMap {
+
+    using MoveVotePair = std::pair<movegen::Move, uint8_t>;
+
+public:
+    MoveVoteMap() = default;
+
+    MoveVoteMap(std::initializer_list<MoveVotePair> init)
+    {
+        assert(init.size() < maxSize);
+
+        for (auto& entry : init) {
+            m_entries.at(m_size++) = std::move(entry);
+        }
+    }
+
+    constexpr void insertOrIncrement(movegen::Move extMove, uint8_t extVote)
+    {
+        bool matchFound = false;
+        for (auto& [move, vote] : *this) {
+            if (move == extMove) {
+                matchFound = true;
+                vote += extVote;
+            }
+        }
+
+        if (!matchFound) {
+            assert(m_size < maxSize);
+            m_entries.at(m_size++) = MoveVotePair(extMove, extVote);
+        }
+    }
+
+    constexpr void clear()
+    {
+        m_size = 0;
+    }
+
+    auto* begin() { return m_entries.begin(); }
+    auto* end() { return m_entries.begin() + m_size; }
+
+    auto* begin() const { return m_entries.begin(); }
+    auto* end() const { return m_entries.begin() + m_size; }
+
+private:
+    std::array<MoveVotePair, maxSize> m_entries {};
+    size_t m_size {};
+};
+
+}
--- a/src/evaluation/searcher.h
+++ b/src/evaluation/searcher.h
@@ -14,6 +14,8 @@
 #include "syzygy/syzygy.h"
 #include <engine/zobrist_hashing.h>
 
+#include <future>
+
 namespace evaluation {
 
 enum SearchType {
@@ -27,6 +29,13 @@ struct MoveResult {
 };
 
 class Searcher;
+
+struct SearcherResult {
+    int32_t score;
+    movegen::Move pvMove;
+    uint8_t searchedDepth;
+    Searcher* searcher;
+};
 
 class Searcher {
 public:
@@ -59,6 +68,36 @@ public:
         m_dtz = 0;
 
         return negamax(depth, board, alpha, beta);
+    }
+
+    void inline startSearchAsync(ThreadPool& threadPool, uint8_t depth, const BitBoard& board, int32_t alpha, int32_t beta)
+    {
+        m_searchPromise = std::promise<SearcherResult> {};
+        m_futureResult = m_searchPromise.get_future();
+
+        m_moveOrdering.pvTable().setIsFollowing(true);
+        m_wdl = syzygy::WdlResultTableNotActive;
+        m_dtz = 0;
+
+        threadPool.submit([this, depth, board, alpha, beta] {
+            SearcherResult result {
+                .score
+                = negamax(depth, board, alpha, beta),
+                .pvMove = m_moveOrdering.pvTable().bestMove(),
+                .searchedDepth = m_moveOrdering.pvTable().size(),
+                .searcher = this
+            };
+
+            /* Stop any searches going on in other threads, so we're not waiting on the slowest search */
+            setSearchStopped(true);
+
+            m_searchPromise.set_value(result);
+        });
+    }
+
+    constexpr std::optional<SearcherResult> getSearchResult()
+    {
+        return m_futureResult.get();
     }
 
     constexpr movegen::Move getPvMove() const
@@ -478,6 +517,9 @@ private:
 
     BitBoard m_board;
     uint64_t m_hash;
+
+    std::promise<SearcherResult> m_searchPromise;
+    std::future<SearcherResult> m_futureResult;
 
     /* search configs */
     constexpr static inline uint32_t s_fullDepthMove { 4 };
--- a/src/movegen/move_types.h
+++ b/src/movegen/move_types.h
@@ -29,7 +29,7 @@ public:
     Move() = default;
 
     explicit Move(uint32_t ext_data)
-        : data(ext_data) { };
+        : data(ext_data) {};
 
     explicit Move(uint64_t from, uint64_t to, uint64_t piece, uint64_t promotion, uint64_t castle, uint64_t capture, uint64_t enPessant, uint64_t takeEnPessant)
         : data(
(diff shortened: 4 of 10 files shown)


src/uci_handler.h
```diff
@@ -175,6 +177,10 @@ class UciHandler {
 
     static bool handleGo(std::string_view args)
     {
+        if (!s_searchersInitialized) {
+            s_evaluator.initializeSearchers(s_numThreads);
+            s_searchersInitialized = true;
+        }
```
hansbinderup, line 183: i don't think that this is the consumer's job.

I think that it would be neat to simply pass amount of threads to the `getBestMove` and it will handle it for us internally :)

src/helpers/game_phases.h
```diff
@@ -65,12 +65,11 @@ constexpr void subPhaseScore(evaluation::gamephase::Score& score, const std::arr
  * using precomputed PESTO piece-square tables.
  * Also the material score (game phase dependent) is increased whenever probed
  */
-constexpr void addPestoPhaseScore(evaluation::gamephase::Score& score, Piece piece, BoardPosition pos)
+constexpr __attribute__((always_inline)) void addPestoPhaseScore(evaluation::gamephase::Score& score, Piece piece, BoardPosition pos)
```
hansbinderup, line 68: I'm confused why this is necessary. constexpr should implicitly be inline. Was this tested with release build as well?

src/evaluation/evaluator.h
```diff
@@ -599,22 +567,365 @@ class Evaluator {
...
+
+                if (timeLeft < timeLimit) {
+                    /* m_logger.log("Stopped early; saved: {}", std::chrono::duration_cast<std::chrono::milliseconds>(timeLeft)); */
+                    break;
+                }
+            }
+
+            /* Thread voting: https://www.chessprogramming.org/Lazy_SMP */
+            std::vector<SearcherResult> searchResults;
+            std::map<movegen::Move, int64_t> movesVotes;
```
hansbinderup, line 836: :wink: *heeeap*

hansbinderup (commented):
Sorry, don't have much time for a better review today. Just some initial thoughts :)

1: the fact that the consumer have to be in charge of initializing the threads etc feels quite error prone. I did some setup on my bench branch and I kept hitting some deadlocks which were pretty annoying. I simply had to "reset" the evaluator and then it worked. I don't think this is the job of the consumer :)

2: the engine loop is now using heap - it should not be necessary :)

src/evaluation/evaluator.h
```diff
@@ -6,322 +6,348 @@
...
-    constexpr void stop()
+    constexpr U& at(const T& ext_key)
     {
-        m_isStopped = true;
+        for (auto& [key, value] : m_entries) {
+            if (key == ext_key) {
+                return value;
+            }
+        }
+        throw std::out_of_range("No entry matching this key");
```
hansbinderup, line 89: We don't catch this anywhere - catch / throw is extremely slow so we should avoid using it

src/evaluation/evaluator.h
```diff
@@ -6,322 +6,348 @@
...
+struct SearcherResult {
+    int32_t score;
+    uint8_t searcherId;
+    movegen::Move pvMove;
+    uint8_t searchedDepth;
+};
+
+// Wrapper around a static array for intuitive iteration
+template<typename T, uint8_t maxSize>
+class StackVector {
```
hansbinderup, line 49: Not the biggest fan of custom containers unless it's something we find ourselves using on a regular basis. 
A simple pre-allocated array and an index can do the same :)

src/evaluation/evaluator.h
```diff
@@ -849,17 +953,19 @@ class Evaluator {
 
                 // If didn't fall out of window and wasn't an immediate early termination, push back to results
                 if ((result.score > alpha) && (result.score < beta) && result.searchedDepth > 0) {
-                    searchResults.push_back(std::move(result));
+                    searchResults.at(searchResults.size()) = std::move(result);
```
hansbinderup, line 968: Why move?

src/evaluation/evaluator.h
```diff
@@ -483,7 +478,7 @@ class Searcher {
 
             undoMove(moveRes->hash);
 
-            if (s_isStopped.load(std::memory_order_relaxed) || m_searchState.load(std::memory_order_relaxed) == SearchState::Kill)
+            if (s_isStopped.load(std::memory_order_relaxed) || s_oneFullSearchComplete.load(std::memory_order_relaxed) || m_searchState.load(std::memory_order_relaxed) == SearchState::Kill)
```
hansbinderup, line 482: This is racy :racing_car: 
Also no need to store 3 atomics when we can just add the stopped / done search to the state :)

src/evaluation/evaluator.h
hansbinderup, line 1: you missed the check in quiesence search as well

src/evaluation/evaluator.h
```diff
@@ -1045,7 +1060,7 @@ class Evaluator {
         timeHandler.detach();
     }
 
-    StackVector<Searcher, s_maxThreads> m_searchers {};
+    StackVector<std::unique_ptr<Searcher>, s_maxThreads> m_searchers {};
```
hansbinderup, line 1066: Why not just use a normal vector? It's allocated on the heap anyways :)

src/evaluation/evaluator.h
```diff
@@ -745,14 +745,17 @@ class Evaluator {
     constexpr movegen::Move
     getBestMove(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
     {
-        m_startTime
-            = std::chrono::system_clock::now();
-
+        {
+            std::unique_lock<std::mutex> lock(m_timeHandleMutex);
```
hansbinderup, line 749: The sanitizer might flag this as racy but I don't think this solves the issue. 
starttime and endtime are still set outside their own locks. Might be "memory safe" now but it's still racy :)

`getBestMove` is not designed to be thread safe. Why would we ever want to perform multiple searches? :)

hansbinderup (commented):
Some more comments :)

I think it's important to try to keep things simple at this stage. And take it one step at a time:

1. Ensure search is working as before (nodes, PV lines etc. when running threads=1).
2. Tests with 2 threads should also be fairly consistent (will most likely be non-deterministic due to the nature of lazy smp).
3. Thread voting etc. is not important until we know for sure that §1 and §2  is working.

Minor comments:

It's quite difficult to review when the commits are not linear :/ 

Tools like Helgrind can be helpful, but it doesn't know the bigger picture :)

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+    movegen::Move pvMove;
+    uint8_t searchedDepth;
+};
 
-        /* we're dependent on the hash table - ensure it's initialized! */
-        if (engine::TtHashTable::getSizeMb() == 0) {
-            engine::TtHashTable::setSizeMb(s_defaultTtHashTableSizeMb);
-        }
+template<uint8_t maxSize>
+class MoveVoteMap {
```
hansbinderup, line 41: I think this belongs in its own file :)
(would also have made rebasing easier)

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+};
 
-            fmt::print("Captures[{}]: ", captures.count());
-            for (const auto& move : captures) {
-                fmt::print("{} [{}]  ", move, m_moveOrdering.moveScore(board, move, 0));
-            }
-            fmt::print("\n\n");
-        }
+/* Base class for searcher */
+class Searcher {
```
hansbinderup, line 103: I think this belongs in its own file :)
(would also have made rebasing easier)

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+    {
+        assert(init.size() < maxSize);
+
+        for (auto& entry : init) {
+            m_entries.at(m_size++) = std::move(entry);
+        }
     }
 
-    constexpr std::optional<movegen::Move> getPonderMove() const
+    constexpr uint8_t& at(const movegen::Move& ext_move)
```
hansbinderup, line 57: If use for lookup only, let's add a const non-reference variant as well

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+    uint8_t searcherId;
+    movegen::Move pvMove;
+    uint8_t searchedDepth;
+};
 
-        /* we're dependent on the hash table - ensure it's initialized! */
-        if (engine::TtHashTable::getSizeMb() == 0) {
-            engine::TtHashTable::setSizeMb(s_defaultTtHashTableSizeMb);
-        }
+template<uint8_t maxSize>
```
hansbinderup, line 40: only used for templating so is safer:
```suggestion
template<size_t maxSize>
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+            m_entries.at(m_size++) = std::move(entry);
+        }
     }
 
-    constexpr std::optional<movegen::Move> getPonderMove() const
+    constexpr uint8_t& at(const movegen::Move& ext_move)
     {
-        if (m_moveOrdering.pvTable().size() > 1) {
-            return m_moveOrdering.pvTable()[1];
+        for (auto& [move, vote] : m_entries) {
```
hansbinderup, line 59: No need to iterate all entries - should iterate using the overloads:
```suggestion
        for (auto& [move, vote] : *this*) {
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
         }
+        assert(false);
 
-        return std::nullopt;
+        return m_entries.at(0).second;
     }
 
-    /* TODO: ensure time handler is stopped as well - not just assumed stopped. This is racy */
-    constexpr void stop()
+    constexpr bool count(const movegen::Move& ext_move)
```
hansbinderup, line 69:
```suggestion
    constexpr bool contains(const movegen::Move& ext_move) const
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
     {
-        m_isStopped = true;
+        for (auto& [move, vote] : m_entries) {
```
hansbinderup, line 71:
```suggestion
        for (const auto& [move, vote] : *this) {
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+        m_entries.at(m_size++) = MoveVotePair(move, vote);
     }
 
-    constexpr void printEvaluation(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
+    constexpr void clear()
     {
-        resetTiming(); /* to reset nodes etc but not tables */
+        for (uint8_t i = 0; i < m_size; i++) {
+            m_entries.at(i) = MoveVotePair {};
+        }
```
hansbinderup, line 89:
```suggestion
        for (auto& entry : *this) {
            entry = MoveVotePair {};
        }
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
+        m_size = 0;
+    }
 
-        uint8_t depth = depthInput.value_or(5);
+    auto* begin() { return m_entries.begin(); }
 
-        fmt::println("");
+    auto* end() { return m_entries.begin() + m_size; }
```
hansbinderup, line 95: For `*this` to work we need to add const iterators as well:

```suggestion
    auto* end() { return m_entries.begin() + m_size; }
    
    auto* begin() const { return m_entries.begin(); }
    auto* end() const { return m_entries.begin() + m_size; }
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
-            for (const auto& move : captures) {
-                fmt::print("{} [{}]  ", move, m_moveOrdering.moveScore(board, move, 0));
-            }
-            fmt::print("\n\n");
-        }
+/* Base class for searcher */
+class Searcher {
+public:
+    Searcher()
+        : m_searcherId(s_numSearchers++)
```
hansbinderup, line 106: I think this is safer (and cleaner):

```suggestion
    Searcher(size_t id)
        : m_searcherId(id)
```

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
 
-        movegen::ValidMoves moves;
-        engine::getAllMoves<movegen::MovePseudoLegal>(board, moves);
-        m_moveOrdering.sortMoves(board, moves, m_ply);
-        const int32_t score = negamax(depth, board);
-        stop();
+    ~Searcher()
+    {
+        s_numSearchers--;
+    }
```
hansbinderup, line 115: with above: no need for this as well

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
     }
 
-    void startTimeHandler()
-    {
-        bool wasStopped = m_isStopped.exchange(false);
-        if (!wasStopped) {
-            return; /* nothing to do here */
-        }
+    static uint8_t s_numSearchers;
+    static uint64_t s_nodes;
```
hansbinderup, line 576: Multiple threads can access this - not safe. 
I think it would be better to not make it static and instead accumulate each node count (avoid raciness and atomics)

src/evaluation/evaluator.h
```diff
@@ -499,11 +458,15 @@ class Evaluator {
         return alpha;
     }
 
+    static std::atomic_bool s_searchStopped;
+    static std::atomic_bool s_killed;
```
hansbinderup, line 462:
```suggestion
    static inline std::atomic_bool s_searchStopped { true };
    static inline std::atomic_bool s_killed { false };
```

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
+    constexpr static inline uint8_t s_razorDeepReductionLimit { 2 };
+
+    constexpr static inline uint8_t s_nullMoveReduction { 2 };
+};
+
+uint8_t Searcher::s_numSearchers { 0 };
+uint64_t Searcher::s_nodes { 0 };
+
+std::atomic_bool Searcher::s_searchStopped { true };
+std::atomic_bool Searcher::s_killed { false };
```
hansbinderup, line 611: No need for these when inlined

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
+        }
+
+        else {
+            for (uint8_t i = m_searchers.size(); i < size; i++) {
+                m_searchers.emplace_back(std::make_unique<Searcher>());
+            }
+        }
+
+        m_searchers.resize(size);
+        m_threadPool.resize(size + 2);
```
hansbinderup, line 633: I think we can simplify this a fair bit (it's non performance critical - will be done from uci options and not engine loop):

```suggestion
        if (size == m_searchers.size()) {
            return;
        }

	m_searchers.clear();
	m_searchers.reserve(size);
	
        for (uint8_t i = m_searchers.size(); i < size; i++) {
            m_searchers.emplace_back(std::make_unique<Searcher>());
        }

	/* add main thread + time handler as well */
        m_threadPool.resize(size + 2);
```

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
+    }
+
+    constexpr uint64_t getNodes() const
```
hansbinderup, line 636: should iterate each searcher and accumulate their nodes

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
+    {
+        return Searcher::getNodes();
+    }
+
+    constexpr movegen::Move
+    getBestMove(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
```
hansbinderup, line 642: Happens when you auto format with invalid closures etc: 
```suggestion
    constexpr movegen::Move getBestMove(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
```

src/movegen/move_types.h
```diff
@@ -67,6 +67,12 @@ class Move {
         return Move(from, to, piece, PromotionNone, CastleNone, takeEnPessant, enPessant, takeEnPessant);
     }
 
+    // For construction of maps in thread voting
+    bool operator<(const Move& other) const
```
hansbinderup, line 71: This is not needed anymore

src/evaluation/evaluator.h
```diff
@@ -1,312 +1,271 @@
...
-        fmt::println("");
+    auto* end() { return m_entries.begin() + m_size; }
 
-        movegen::ValidMoves captures;
-        engine::getAllMoves<movegen::MoveCapture>(board, captures);
-        if (captures.count()) {
-            m_moveOrdering.sortMoves(board, captures, m_ply);
+private:
+    std::array<MoveVotePair, maxSize> m_entries {};
+    uint8_t m_size {};
```
hansbinderup, line 99:
```suggestion
    size_t m_size {};
```

hansbinderup (commented):
It would have been a whole lot easier to rebase if the new features were added to new files :) 

A few comments - mostly optimizations and code suggestions :)

src/evaluation/evaluator.h
clarashepherd: The only use of `at` is in the voting, and we want the value (counter) to be mutable. So I think the signature is correct here?

src/evaluation/evaluator.h
clarashepherd: No longer needed, because I've removed searcher ID entirely.

src/evaluation/evaluator.h
clarashepherd: No longer needed, because I've removed searcher ID entirely.

src/evaluation/evaluator.h
clarashepherd: Sorted :) Now the evaluator owns the its own m_nodes and the evaluator does summation on `printScoreInfo`, which it now owns.

src/evaluation/evaluator.h
clarashepherd: See above :)

src/evaluation/evaluator.h
clarashepherd: How is this a simplification? Do you mean in terms of understand the execution?

Are there not any states we want to keep persistent when the number of threads is changed?

src/evaluation/evaluator.h
clarashepherd: Done.

src/evaluation/evaluator.h
clarashepherd: Done. It's handy to know for the future that a tricky rebase can be a hint towards file seperation :)

src/evaluation/evaluator.h
clarashepherd: Done.

src/evaluation/evaluator.h
clarashepherd: This is slick! Done and thanks :)

src/evaluation/evaluator.h
clarashepherd: Done.

src/evaluation/evaluator.h
clarashepherd: Again, this is nice to know about - thanks

src/evaluation/evaluator.h
clarashepherd: Done :)

src/evaluation/evaluator.h
clarashepherd: Done.

src/movegen/move_types.h
clarashepherd: Removed.

src/evaluation/evaluator.h
clarashepherd: Done.

src/evaluation/evaluator.h
```diff
@@ -614,43 +572,428 @@ class Evaluator {
...
+                const auto move = result.pvMove;
+
+                /* Can be tweaked for optimization */
+                int64_t voteWeight = (score - s_minScore) * depth;
+
+                if (!m_movesVotes.count(move)) {
+                    m_movesVotes.insert(move, voteWeight);
+                } else {
+                    m_movesVotes.at(move) += voteWeight;
+                }
```
hansbinderup, line 928: Can be optimized with ["insertOrAssign"](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) :)

src/evaluation/evaluator.h
hansbinderup: The thread size is only set at start and should, in theory, never be changed "runtime".
In any case, we do not want to persist any state. If it's done mid game, it will be super fucked anyways ^^

Therefore I think it's important to just keep it as simple as possible and not try to "optimize for something that won't happen" :)

src/evaluation/evaluator.h
```diff
@@ -16,9 +17,34 @@ namespace evaluation {
...
+
+        m_searchers.resize(size);
+        m_threadPool.resize(size + 2);
+    }
+
     constexpr uint64_t getNodes() const
     {
-        return m_searcher.getNodes();
+        uint64_t totalNodes {};
+        for (auto& searcher : m_searchers) {
```
hansbinderup, line 44:
```suggestion
        for (const auto& searcher : m_searchers) {
```

src/evaluation/evaluator.h
```diff
@@ -33,22 +59,26 @@ class Evaluator {
...
         if (m_isPondering || !m_ponderingEnabled)
             return false;
 
         m_isPondering = true;
-        return getBestMoveAsync(board);
+        return getBestMoveAsync(board, numSearchers);
     }
 
-    constexpr bool getBestMoveAsync(const BitBoard& board, std::optional<uint8_t> depthInput = std::nullopt)
+    constexpr bool getBestMoveAsync(const BitBoard& board, uint8_t numSearchers, std::optional<uint8_t> depthInput = std::nullopt)
```
hansbinderup, line 78: I don't think it's necessary to pass the searchers as an argument.
We had a talk about it and the prior implementation (your first draft) it was the consumers responsibility to resize, check etc.

I think the searchers should only be resized from the uci option which will happen at the start and nowhere else :)

src/evaluation/evaluator.h
```diff
@@ -57,9 +87,8 @@ class Evaluator {
 
             fmt::print("bestmove {}", move);
             if (m_ponderingEnabled) {
-                const auto ponderMove = m_searcher.getPonderMove();
-                if (ponderMove.has_value()) {
-                    fmt::print(" ponder {}", ponderMove.value());
+                if (m_ponderMove.has_value()) {
```
hansbinderup, line 90: rebase conflict

src/evaluation/evaluator.h
```diff
@@ -183,15 +302,20 @@ class Evaluator {
...
 
-    ThreadPool m_threadPool { 1 }; /* single searcher */
+    ThreadPool m_threadPool { 3 }; /* Iterative deepening, time handler, default=1 searcher */
+
+    std::vector<std::unique_ptr<Searcher>> m_searchers {};
+    MoveVoteMap<s_maxThreads> m_movesVotes {};
+
     bool m_isPondering { false };
     bool m_ponderingEnabled { false };
+    std::optional<movegen::Move> m_ponderMove {};
```
hansbinderup, line 318: rebase conflict

src/evaluation/move_vote_map.h
```diff
@@ -0,0 +1,59 @@
...
+    MoveVoteMap(std::initializer_list<MoveVotePair> init)
+    {
+        assert(init.size() < maxSize);
+
+        for (auto& entry : init) {
+            m_entries.at(m_size++) = std::move(entry);
+        }
+    }
+
+    constexpr void insertOrIncrement(movegen::Move ext_move, uint8_t ext_vote)
```
hansbinderup, line 24: We don't use snake case :wink: 
```suggestion
    constexpr void insertOrIncrement(movegen::Move extMove, uint8_t extVote)
```

src/evaluation/move_vote_map.h
```diff
@@ -0,0 +1,59 @@
...
+            assert(m_size < maxSize);
+            m_entries.at(m_size++) = MoveVotePair(ext_move, ext_vote);
+        }
+    }
+
+    constexpr void clear()
+    {
+        for (auto& entry : *this) {
+            entry = MoveVotePair {};
+        }
```
hansbinderup, line 44: this is, in theory, not necessary. 
I don't have strong opinions about it, but it *should* be enough to just reset the size :)

src/evaluation/searcher.h
```diff
@@ -61,6 +70,36 @@ class Searcher {
...
+            };
+
+            setSearchStopped(true);
+
+            m_searchPromise.set_value(result);
+        });
+    }
+
+    constexpr std::optional<SearcherResult>
+    getSearchResult()
```
hansbinderup, line 98: line break

src/evaluation/searcher.h
```diff
@@ -61,6 +70,36 @@ class Searcher {
...
+        threadPool.submit([this, depth, board, alpha, beta] {
+            SearcherResult result {
+                .score
+                = negamax(depth, board, alpha, beta),
+                .pvMove = m_moveOrdering.pvTable().bestMove(),
+                .searchedDepth = m_moveOrdering.pvTable().size(),
+                .searcher = this
+            };
+
+            setSearchStopped(true);
```
hansbinderup, line 91: A comment here is probably nice for the future

src/uci_handler.h
```diff
@@ -369,5 +370,8 @@ class UciHandler {
         ucioption::make<ucioption::check>("Ponder", false, [](bool enabled) { s_evaluator.setPondering(enabled); }),
         ucioption::make<ucioption::string>("Syzygy", "<empty>", syzygyCallback),
         ucioption::make<ucioption::spin>("Hash", s_defaultTtHashTableSizeMb, ucioption::Limits { .min = 1, .max = 1024 }, [](uint64_t val) { engine::TtHashTable::setSizeMb(val); }),
+        ucioption::make<ucioption::spin>("Threads", 1, ucioption::Limits { .min = 1, .max = 256 }, [](uint64_t val) {
```
hansbinderup, line 373: I think that this should set the searchers directly. Then it's no need to store a local copy.
Also with https://github.com/hansbinderup/meltdown-chess-engine/pull/72 it's not necessary to pass to bench :)

hansbinderup (commented):
A few minor comments. Looks super :1st_place_medal: 

Just as a reference: [OpenBench result](https://openbench.bunny.beer/test/67/)

Really happy for you putting in the work for this! :pray:

src/evaluation/move_vote_map.h
clarashepherd: I added this for cleanness, because it makes debugging etc easier. But since the voting map isn't complicated right now, I agree that we can remove it.

src/evaluation/evaluator.h
clarashepherd: I don't think this is just a rebase conflict. `scanForBestMove` does thread voting and returns only returns the best move. From the 'winning' result the ponder move is also needed by `getBestMove`.

We could leave this as is, change the signatures, or secret third thing. What do you think?

hansbinderup: Agree, I think the current solution is fine. Sorry, was a little too fast there :)

hansbinderup (approved):
Perfect, let's get this in!

src/evaluation/evaluator.h
```diff
@@ -218,7 +216,9 @@ class Evaluator {
                 alpha = score - s_aspirationWindow;
                 beta = score + s_aspirationWindow;
 
-                printScoreInfo(singleSearcher.get(), board, score, d);
+                if (TimeManager::timeForAnotherSearch)
+                    printScoreInfo(singleSearcher.get(), board, score, d);
```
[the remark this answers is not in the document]
clarashepherd: Good spot! Removed the 'if'

PR #48, opened 2025-04-09
uci_handler: add uci options generalization

hansbinderup:
Previously it was quite a pain to add new UCI options. You had to add it manually to the "uci ok string", the debug handle and manually parse all the data.

Now instead, all you have to do is add it to the list and implement the callback.

Adding new types is also easy but requires a bit more work.

This implementation should be better, safer and way faster for new additions.

### DIFF
--- a/src/uci_options.h
+++ b/src/uci_options.h
@@ -1,0 +1,149 @@
+#pragma once
+
+#include "parsing/input_parsing.h"
+
+#include "fmt/base.h"
+#include <cstdint>
+#include <functional>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace ucioption {
+
+/* uci types for easier creation */
+using string = std::string;
+using check = bool;
+using spin = uint64_t;
+
+template<typename T>
+struct Storage {
+    T value;
+    const T defaultValue;
+    const std::function<void(const T&)> callback;
+};
+
+using Variant = std::variant<
+    Storage<std::string>,
+    Storage<bool>,
+    Storage<uint64_t>>;
+
+struct Limits {
+    uint64_t min;
+    uint64_t max;
+};
+
+struct UciOption {
+    const std::string_view name;
+    Variant variant;
+    const std::optional<Limits> limits {};
+};
+
+template<typename T>
+constexpr auto make(std::string_view name, T defaultVal, std::function<void(const T&)> cb)
+{
+    return UciOption {
+        .name = name,
+        .variant = Storage<T> { .value = defaultVal, .defaultValue = defaultVal, .callback = std::move(cb) },
+    };
+}
+
+template<typename T>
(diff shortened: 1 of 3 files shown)


clarashepherd (approved):

src/uci_options.h
```diff
@@ -0,0 +1,159 @@
...
+using spin = uint64_t;
+
+template<typename T>
+struct Storage {
+    T value;
+    const T defaultValue;
+    const std::function<void(const T&)> callback;
+};
+
+using Variant = std::variant<
```
clarashepherd, line 27: I really like this as a solution to class inheritance plus dynamic sizing, without heap allocation :)

src/uci_options.h
```diff
@@ -0,0 +1,159 @@
...
+                arg.value = false;
+            } else {
+                error = true;
+            }
+        } else if constexpr (std::is_same_v<T, Storage<std::string>>) {
+            arg.value = input;
+        } else if constexpr (std::is_same_v<T, Storage<uint64_t>>) {
+            const auto inputNum = parsing::to_number(input);
+            if (!inputNum.has_value())
+                return false;
```
clarashepherd, line 81: Possible nitpick: why is there a mix of returning false directly, and setting `error` to false before returning it with no handling? The former seems less prone to accidentally setting data in the error case.

src/uci_options.h
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        if (error) {
+            return error;
+        }
+
+        if (arg.callback) {
+            arg.callback(arg.value);
+        } else {
+            fmt::println("Callback is not set for {}", name);
```
clarashepherd, line 101: Should this be a static assert, in line with the above? This is a message for developers rather than users.

clarashepherd (approved):

src/uci_options.h
hansbinderup: It's pretty cool, yes! Can lead to some pretty complicated implementations if you don't take care though.

src/uci_options.h
hansbinderup: Good catch, I think this is leftovers that I forgot to remove. Let me remove it :+1:

src/uci_options.h
hansbinderup: The callback is not a constexpr, so it can not be evaluated at compile time.
I will change it to a runtime assert so we can handle it on debug level instead.

Just wanted a clear error here as it might be difficult to debug if something were to go wrong. Shouldn't happen though (famous last words..)

src/uci_options.h
hansbinderup: Just a heads-up in regards to `dynamic sizing`:

This size allocated is the biggest of the variants. So it's not actually memory efficient if you were to have bigger types. For our sake it really doesn't matter though. But something to be aware of :)

PR #49, opened 2025-04-13
terms: correct pawn shield scores

hansbinderup:
Previously the shield was calculated as any piece in the attack mask of
the king. This is not necessarily a good indicator for shielding. The
queen could eg. be the shield.

Instead turn the score into a pawn shield: pawns in front of the king
where the score is given based on how far the pawn is from the king.

Signed-off-by: Hans Binderup <hbinderup94@gmail.com>

NOTE: currently targeted the tuner branch #46. Should be merged after the tuner and included on main instead.


hansbinderup:
@clarashepherd I've rebased and it should be ready for review :)
Currently running [bench](https://openbench.bunny.beer/test/103/)

clarashepherd (approved):

PR #50, opened 2025-04-13
uci: add Hash option

hansbinderup:
Previously the TT hash table was a fixed size.
This commit adds dynamic scaling of the table.

The option has been added to the UCI handler so it can be provided as an
UCI option.

Due to the atomics constructors I decided to implement our own
allocator. It seemed simpler than trying to make it work with a std
container.

Signed-off-by: Hans Binderup <hbinderup94@gmail.com>

NOTE: should be merged after #48


src/engine/tt_hash_table.h
```diff
@@ -106,13 +106,27 @@ struct TtHashEntry {
...
     {
+        if (s_ttHashSize > 0) {
+            std::free(s_ttHashTable);
+        }
 
-        for (auto& entry : s_ttHashTable) {
-            entry.key = 0;
-            entry.data = TtHashEntryData();
+        s_ttHashSize = tableSizeFromMb(sizeMb);
+        s_ttHashTable = static_cast<TtHashEntry*>(std::aligned_alloc(alignof(TtHashEntry), s_ttHashSize * sizeof(TtHashEntry)));
```
clarashepherd, line 118: Are we making a considered exception to our 'no heap allocation' guideline here? I think it makes sense, but I'd like to check we're still on the same page about the guidelines :)

src/engine/tt_hash_table.h
```diff
@@ -198,12 +216,16 @@ class TtHashTable {
         entry.data.store(newData, std::memory_order_relaxed);
     }
 
+    constexpr static std::size_t tableSizeFromMb(size_t sizeMb)
+    {
+        return (sizeMb * 1024 * 1024) / sizeof(TtHashEntry);
+    }
+
 private:
-    constexpr static std::size_t s_hashTableSizeMb { 128 };
```
clarashepherd, line 202: Why are we changing the default size by a factor of 8?

src/engine/tt_hash_table.h
hansbinderup: For the hash table we don't have much of a choice. I has to be scalable (tournament rules etc).
But for the sake of performance, it should really not matter here.
The guidelines are for the engine itself (searcher) where runtime heap allocations will be super slow compared to stack allocations :)

src/engine/tt_hash_table.h
hansbinderup: Most engines use 16 as default so it seems to be the better practice. 
Also 128 is actually fairly big ^^

hansbinderup:
Windows builds are acting up a bit here. Need to do some tests on a windows PC before merging.

hansbinderup:
Just tested this on my partner's windows PC. It seems to work :+1:

PR #51, opened 2025-04-14
engine: add bench

hansbinderup:
This commit adds the option to "bench" the engine. This is a "more or less standardized" way of benchmarking engines.

The positions are used by most engines out there.

We have two ways of running a bench:
From the UCI handler: bench <depth>
From cmd line directly: ./meltdown bench

The latter is necessary for eg. openbench.

### DIFF
--- a/src/engine/bench.h
+++ b/src/engine/bench.h
@@ -1,0 +1,115 @@
+#pragma once
+
+#include "evaluation/evaluator.h"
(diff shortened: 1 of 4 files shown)


clarashepherd (commented):
Awesome, looking forward to giving this a go :dancers:

clarashepherd (approved):
Awesome, looking forward to giving this a go :dancers:

PR #63, opened 2025-04-26
thread_pool: add thread pool and use it in the evaluator

hansbinderup:
See each commit :)


hansbinderup:
Changed the base from the feature branch -> main. So this PR also contains the single commit from the feature branch (enable PR workflow for feature branches - sorry about that :) )

clarashepherd (approved):

PR #66, opened 2025-04-30
tests: improve local tournament scripts

hansbinderup:
During the blackout in Spain is was forced to run a lot of local tests instead of using OpenBench.. It can still be nice to run some initial tests locally, so I thought that it's nice to have an easier time doing that.

Still room for some improvement here, but should be a nice little upgrade for now.
Remember to update the config file after pulling these commits! :)


clarashepherd (approved):
So nice! Approved :sun_with_face:

PR #67, opened 2025-04-30
time_manager: initial time manager

hansbinderup:

### DIFF
--- a/src/evaluation/searcher.h
+++ b/src/evaluation/searcher.h
@@ -47,6 +47,11 @@ public:
         return m_selDepth;
     }
 
+    static constexpr void setSearchStopped(bool value)
+    {
+        s_searchStopped.store(value, std::memory_order_relaxed);
+    }
+
     int32_t inline startSearch(uint8_t depth, const BitBoard& board, int32_t alpha, int32_t beta)
     {
         m_moveOrdering.pvTable().setIsFollowing(true);
@@ -300,7 +305,7 @@ public:
 
             undoMove(moveRes->hash);
 
-            if (TimeManager::hasTimedOut())
+            if (isSearchStopped())
                 return score;
 
             if (score >= beta) {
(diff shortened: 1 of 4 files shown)


src/time_manager.h
```diff
@@ -0,0 +1,169 @@
...
+        s_startTime = std::chrono::system_clock::now();
+        s_endTime = TimePoint::max();
+        s_timedOut = false;
+    }
+
+    /* should only be called from searchers
+     * will check for timeout from the searchers themselves
+     * this will remove the need for a time thread or similar
+     * and will make the engine more deterministic */
+    static inline bool searcherCheckForTimeout(uint64_t nodes)
```
clarashepherd, line 52: If this is only to be called from the searchers, does it not belong in the searcher class?

src/evaluation/searcher.h
```diff
@@ -454,8 +449,6 @@ class Searcher {
 
     static inline uint8_t s_numSearchers {};
 
-    static inline std::atomic_bool s_searchStopped { true };
```
clarashepherd, line 457: This flag will probably have to be re-introduced for lazySMP, because searches can be stopped for reasons other than a timeout.

I think there's a discussion to be had here about what's cleanest - adding and removing components, or leaving them in for fewer changes across the history.

clarashepherd (commented):
Looks great! Thanks for clearing this up, especially for removing raciness from the time handler :palm_tree: 

Left a few stylistic questions that don't necessarily merit changes - maybe just a discussion.

src/time_manager.h
hansbinderup: I agree :+1: will move the check to the searcher instead :)
Was trying to have minimal impact on the searcher when I implemented this initially (to avoid rebase chaos) :)

src/evaluation/searcher.h
hansbinderup: With the change from above, I can reintroduce the flag.
In general I prefer to have each commit being independent. They should always compile and not have logical errors.
That said I think this is a minor, so I'll just introduce it in the check from above - then it should play nicely with your branch :)

clarashepherd (approved):

PR #72, opened 2025-05-02
bench: pass evaluator instead of self ownning

hansbinderup:
Now that we have thread pools we don't need to allocate threads just for a "debug/testing handle".
Instead we can just pass an evaluator and let the consumer have ownership instead.

This has a few extra benefits:
1. No extra threads running
2. Thread count will automatically be updated (no need to pass args etc)
3. Less memory allocated

Bench 11349788

### DIFF
--- a/src/engine/bench.h
+++ b/src/engine/bench.h
@@ -11,11 +11,9 @@ namespace engine {
 
 class Bench {
 public:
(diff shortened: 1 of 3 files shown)


clarashepherd (approved):
Thanks for handling the rebase of this :)

hansbinderup:
> Thanks for handling the rebase of this :)

Thanks for handling the other 500 rebases :)

PR #77, opened 2025-05-06
uci_handler: correct threads option

hansbinderup:
This commit fixes overflows and incorrect buffer size for searchers.

Prior to this commit it was possible to add "256" threads; uint8_t would overflow and the search result array was limited to 128 (local constant).

Bench 11029400


clarashepherd:
Oops, went through one iteration too many with the handling of the threads option during the monolithic lazySMP PR.

Thanks for the fixup :sparkling_heart:

clarashepherd (approved):

PR #81, opened 2025-05-08
movegen: scale moves to fit in 16bits instead of 32bit

hansbinderup:
We used to store quite a bit of metadata in the moves. For this to work we had to allocate 32bits for each move.
If we move the "piece" meta data to a look-up we can instead half the memory and store almost the same information.
This is beneficial for TT as we'll create 16bits of data that we can use for something else (todo).

Bench 12149207

[Test](https://openbench.bunny.beer/test/106/) is running


clarashepherd (approved):
Great spot and optimization :star_struck:

PR #83, opened 2025-05-09
docker: Add clang to Dockerfile

clarashepherd:
Tuner requires clang-format, which previously needed to be installed manually. Not any more...

Bench 8880290


hansbinderup (commented):
Remember to update the github workflows as well :)

hansbinderup (approved):
Looks good :+1: 

Just a note on the commits; I don't think it's necessary to divide it into multiple commits. It's easier to manage if we can just eg. revert one commit and then it still works. With 2 commits they depend on each other - but in a bad way :)

PR #85, opened 2025-05-11
searcher: add internal iterative reductions

hansbinderup:
Internal Iterative Reductions (IIR) is a newer and more commonly used variant of Internal Iterative Deepening (IID).

The idea is that when no hash move is found, the guess is that the node must not be very important (at the current iteration).

Instead search one shallower depth and let the next iteration handle the deeper level (if found more important at this point).

Bench 8385565

[Test](https://openbench.bunny.beer/test/130/) is running

### DIFF
--- a/src/evaluation/searcher.h
+++ b/src/evaluation/searcher.h
@@ -193,7 +193,7 @@ public:
     }
 
     template<SearchType searchType = SearchType::Default>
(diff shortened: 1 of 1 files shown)


clarashepherd (approved):

clarashepherd:
I've provisionally approved, but I think we should run more tests - seems neutral right now.

hansbinderup:
> I've provisionally approved, but I think we should run more tests - seems neutral right now.

I've added it to the SPSA tuning branch and will wait until we have a tuning value etc :)

PR #91, opened 2025-05-12
searcher: add search stack

hansbinderup:
This commit adds a search stack for the searchers. Previously we stored a local hash copy + "new board" when performing the search recursion.

It is beneficial to actually store all the positions we've been through during the search. This means that we can, in theory, iterate the stack and get a better understanding on how the current search is performing.

Furthermore it should improve the cache as we no longer allocate the temporary hash and board on the stack.

Bench 8741465

```
Elo   | 2.53 +- 8.02 (95%)
LLR   | 0.67 (-2.94, 2.94) [-5.00, 0.00]
Games | N: 2330 W: 763 L: 746 D: 821
Penta | [48, 172, 711, 183, 51]
https://openbench.bunny.beer/test/145/
```

### DIFF
--- a/src/evaluation/searcher.h
+++ b/src/evaluation/searcher.h
@@ -43,7 +43,7 @@ public:
 
     void setHashKey(uint64_t hash)
     {
-        m_hash = hash;
+        m_stack.front().hash = hash;
     }
 
     constexpr uint64_t getNodes() const
@@ -63,6 +63,8 @@ public:
 
     Score inline startSearch(uint8_t depth, const BitBoard& board, Score alpha, Score beta)
     {
+        assert(m_stackItr == m_stack.begin());
+
         m_moveOrdering.pvTable().setIsFollowing(true);
         m_wdl = syzygy::WdlResultTableNotActive;
         m_dtz = 0;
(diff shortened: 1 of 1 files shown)


src/evaluation/searcher.h
```diff
@@ -520,8 +532,13 @@ class Searcher {
     syzygy::WdlResult m_wdl { syzygy::WdlResultTableNotActive };
     uint8_t m_dtz {};
 
-    BitBoard m_board;
-    uint64_t m_hash;
+    struct StackInfo {
```
clarashepherd, line 535: Is it conventional to define a struct within a class?

I'm not used to seeing this, and we're already namespacing the other structs used by the searcher.

I wonder whether we should consistently move helper structs either inside or outside. What do you think?

hansbinderup: Depends on ownership. In this case the type belongs to the class why it's defined there.
Namespaces expose/group types - classes _can_ do the same. 

Notice that the struct is defined privately meaning that the type can only be used by the class itself :)

clarashepherd:
RE your testing: have you stress tested this for scenarios where the stack will be most memory-heavy? I imagine that would be deep searches in long games.

src/evaluation/searcher.h
clarashepherd: I like the explicitness - thanks for clarifying :)

hansbinderup:
> RE your testing: have you stress tested this for scenarios where the stack will be most memory-heavy? I imagine that would be deep searches in long games.

The stack is pre-allocated so it should not have any impact :) 

The edge case you're thinking of might be endgame searches where we reach the search depth limit?

clarashepherd (approved):

PR #94, opened 2025-05-14
searcher: add tt probe to qsearch

hansbinderup:
As we've added more information to the TT (eval) we might as well use that information in our qsearch as well.

This commit adds the same logic, as used in negamax, to the qsearch, where we always try to use the TT entries if possible. We'll also start updating the evaluation, if allowed, from the qsearch, as it will most likely be a position we'll encounter again in future searches.

I've generalized the extraction of tt moves and tt evaluation to avoid having the same code in both places.

Bench 7987248

[Test](https://openbench.bunny.beer/test/146/) is running


clarashepherd (approved):
Looks great, no notes :)

PR #95, opened 2025-05-14
workflows: discontinue nightly

hansbinderup:
Looking at our release page it's clear that the nightly build is heavily polluting the list.
Further more, the nightlies are not even tagged meaning that they're basically useless in terms of tracking regressions. Instead I've updated the server scripts to checkout main and perform a local build instead.

We won't have nightly testing of the CI builds, but we'll still have nightly versions running on the server.

Bench 8578590


hansbinderup:
FYI @clarashepherd - I'm gonna delete all the nightlies in the release list. They're not tagged, so they're basically useless. Just polluting our release list :)

PR #99, opened 2025-05-18
terms: add term context

hansbinderup:
Instead of computing the same values multiple times we can just store them locally and pass them to our term methods.

An example of this is the pawn attacks which are being computed in multiple terms - we can reuse those values and speed up the evaluation.

This is both a speed up but also a clean up. We can now streamline the methods and make the function signatures a bit prettier.

Bench 8564539

[Test](https://openbench.bunny.beer/test/181/) is running


src/evaluation/term_methods.h
```diff
@@ -38,6 +37,13 @@
 
 namespace evaluation {
 
+/* A context to precompute heavy operations so we don't have to do that
+ * multiple times for a single evaluation
+ * NOTE: Only values that are being reused between terms should be added here */
+struct TermContext {
+    const std::array<uint64_t, magic_enum::enum_count<Player>()> pawnAttacks;
```
clarashepherd, line 44: Is the inclusion of only pawn attacks an intuitive one, or did you try including some other heavy operations?

clarashepherd (commented):
Looks great :) I added one query about how far we can take this, and whether it makes sense to do that now.

src/evaluation/term_methods.h
hansbinderup: Current we only do non-position based computations with the pawn attacks. The other computations are based on the pieces positions, so those we can't precompute. But good observation :+1:

hansbinderup:
Merging as https://github.com/hansbinderup/meltdown-chess-engine/pull/100 is approved

PR #100, opened 2025-05-18
terms: add pawn attacks to mobility scores

hansbinderup:
This commit utilizes the newly added "TermContext". We can now bring
pawn attacks into other terms with literally no cost.

This means that we can improve our mobility terms; only apply the score
if the square is not attacked by opponent's pawns.

Also it fixes a minor bug where we included our own pieces' squares in
the mobility. This is of course not valid, so we should not include
that.

```
Elo   | 14.67 +- 8.24 (95%)
LLR   | 2.94 (-2.94, 2.94) [0.00, 5.00]
Games | N: 4646 W: 1706 L: 1510 D: 1430
Penta | [233, 477, 776, 535, 302]
https://openbench.bunny.beer/test/187/
```

Bench 8444587

#99 should be merged first.


clarashepherd (approved):
Looks good, approved :)

PR #101, opened 2025-05-19
searcher: add lmr table

hansbinderup:
This commit adds dynamic scaling of the Late Move reductions. Previously we had a fixed reduction (2) which was a fine starting point.

With this commit we scale the reduction based on two parameters: depth and amount of moves already searched.

This means that the deeper we're into a search and the more move's we've searched at that depth the higher reduction.

This is quite huge which is also reflected in the bench nodes. We're pruning a lot now!

The table values are borrowed from the engine Carp - a lot of engines use these values these days, so it's probably not a bad starting point :)

Bench 5050092


src/evaluation/searcher.h
```diff
@@ -316,20 +317,25 @@ class Searcher {
...
                 if (movesSearched >= spsa::fullDepthMove
-                    && depth >= spsa::lmrReductionLimit
-                    && !isChecked
                     && !move.isCapture()
                     && !move.isPromotionMove()) {
+
+                    const bool isGivingCheck = engine::isKingAttacked(m_stackItr->board);
+                    int8_t lmrReduction = spsa::lmrBaseReduction + getLmrReduction(depth, movesSearched);
+
+                    lmrReduction -= isChecked; /* reduce less when checked */
```
clarashepherd, line 332: Implicit conversion from bool to int here. What do you think about that?

hansbinderup: Made it explicit - you're right, it's better :+1:

src/evaluation/searcher.h
```diff
@@ -316,20 +317,25 @@ class Searcher {
...
-             * LMR
-             * https://wiki.sharewiz.net/doku.php?id=chess:programming:late_move_reduction
-             */
+            /* Late Move Reduction (LMR)
+             * https://wiki.sharewiz.net/doku.php?id=chess:programming:late_move_reduction */
             if (movesSearched == 0) {
                 score = -negamax(depth - 1, m_stackItr->board, -beta, -alpha);
             } else {
                 if (movesSearched >= spsa::fullDepthMove
-                    && depth >= spsa::lmrReductionLimit
```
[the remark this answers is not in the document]
hansbinderup: It's included in new lmr table instead. Before it was a fixed value, now it will be reflected in the table result. The table can return 0 as well :)

PR #102, opened 2025-05-20
searcher: depth based NMP | zugzwang check for NMP

hansbinderup:
Previously we would reduce depth with a fixed value when doing null
move pruning. We'd also use a naive depth check.

Instead replace the depth check with an evaluation margin check and
prune based on our current depth.

This allows us to do more fancy tricks in the future as well (improving
heuristics eg.)

Bench 7935208

```
zugswang only:
Elo   | 8.86 +- 6.10 (95%)
LLR   | 2.95 (-2.94, 2.94) [-5.00, 0.00]
Games | N: 3882 W: 1261 L: 1162 D: 1459
Penta | [55, 314, 1125, 371, 76]
https://openbench.bunny.beer/test/199/

combined:
Elo   | 24.81 +- 11.55 (95%)
LLR   | 3.03 (-2.94, 2.94) [-5.00, 0.00]
Games | N: 2160 W: 727 L: 573 D: 860
Penta | [82, 223, 358, 293, 124]
https://openbench.bunny.beer/test/200/
```

### DIFF
--- a/src/bit_board.h
+++ b/src/bit_board.h
@@ -110,6 +110,17 @@ struct BitBoard {
         return (attacks[PlayerWhite] & occupation[PlayerBlack]) == 0 && (attacks[PlayerBlack] & occupation[PlayerWhite]) == 0;
     }
 
+    /* this is a very primitive way of checking for zugzwang position
+     * zugzwang is _very rare_ to happen outside of these conditions
+     * so for most cases this should be _super good enough_ */
+    constexpr inline bool hasZugzwangProneMaterial() const
+    {
+        if (player == PlayerWhite)
+            return occupation[PlayerWhite] == (pieces[WhitePawn] | pieces[WhiteKing]);
+        else
+            return occupation[PlayerBlack] == (pieces[BlackPawn] | pieces[BlackKing]);
+    }
+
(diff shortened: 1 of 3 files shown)


clarashepherd (commented):
I'm a little suspicious of these Elo changes. The commit for depth-based reduction seems to lead to more pruning and a reduced bench, while the zugzwang commit seems to increase it again.

I wonder whether there could be a bug in the latter that we're not seeing due to Elo changes cancelling one another out. Have you tried testing each commit individually?

hansbinderup:
> I'm a little suspicious of these Elo changes. The commit for depth-based reduction seems to lead to more pruning and a reduced bench, while the zugzwang commit seems to increase it again.
> 
> I wonder whether there could be a bug in the latter that we're not seeing due to Elo changes cancelling one another out. Have you tried testing each commit individually?

It was my mistake during testing. The Zugzwang position is only happening in endgames. So testing it with a non-endgame book did not make sense. Or it did, to some extend - we don't gain elo in the early/middle game.

With the NMP changes it will be beneficial to check for Zugzwang, as we prune even deeper there. So it would be absolutely devastating to do it in Zugzwang positions :)

PR #106, opened 2025-05-21
move_ordering: centralize scoring/following

clarashepherd:
This flag was redundant, because it was guaranteed to always equal `m_isFollowing` at the time of move sorting.

Bench 4297313

https://openbench.bunny.beer/test/207/


clarashepherd:
Closing for now, because `m_isFollowing` differs from `m_isScoring` at the root, and anyway the move picker is due for a refactor.

PR #118, opened 2025-05-30
meson: compile static linux binaries 

### LINKED ISSUE #113: Compile v1.6 on Linux - solved
I discovered i can only compile with Meson when i edit `/targets/linux[..].txt` like this :

```
[binaries]
c = 'x86_64-linux-gnu-gcc-13'
cpp = 'x86_64-linux-gnu-g++-13'
(...)
```

As you see, i added "-13" to suit a proper compiler version.

[ i'm Xubuntu 22.04 ]

hansbinderup:
Previously we would only provide the static flags for windows executables.
But as some Linux users with different OS and gcc versions are unable to
run the executables then ensure that we provide a somewhat static bundle.

Bench 4247247

closes #113

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,15 +30,18 @@ fmt_dep = fmt.get_variable('fmt_dep')
 fathom = subproject('fathom')
 fathom_dep = fathom.get_variable('fathom_dep')
 
-# extra steps are required for windows static binaries
-if host_machine.system() == 'windows'
-   add_project_link_arguments('-static', '-static-libgcc', '-static-libstdc++',  language: 'cpp')
-endif
-
 meltdown_dep = declare_dependency(
   include_directories: 'src',
   dependencies: [magic_enum_dep, fmt_dep, fathom_dep]
 )
+
+# statically link the C++ standard library and libgcc to avoid runtime issues
+# use full static linking (-static) only on Windows to avoid DLL dependencies
+if host_machine.system() == 'windows'
+  add_project_link_arguments('-static', '-static-libgcc', '-static-libstdc++', language: 'cpp')
+else
+  add_project_link_arguments('-static-libgcc', '-static-libstdc++', language: 'cpp')
+endif
 
 # build tuner instead of meltdown engine
 if get_option('tuning') == true


tissatussa:
I tried to compile by Meson (./scripts/build.sh --release) and by Docker (./docker/run.sh and then ./scripts/build.sh --release) but both failed. Remember i don't want to change my default GCC version (i have v11, see #113). Again, i managed to compile with Meson (v1.8.0) by adding "-13" to the GCC lines in ./targets/*.txt .. although you discouraged this 'hack', this seems most easy and logical to me.

hansbinderup:
Hi @tissatussa, thanks for the update! 

> I tried to compile by Meson (./scripts/build.sh --release) and by Docker (./docker/run.sh and then ./scripts/build.sh --release) but both failed. Remember i don't want to change my default GCC version (i have v11, see #113). 

I think I just managed to reproduce. 
Problem is that you already have a meson project configured from your OS user. The docker container then tries to reconfigure and does not have permissions.

I'll fix this. But for now you can do this:

```
# remove build cache
[hans@archlinux meltdown-chess-engine]$ rm -rf .build

# mount docker container
[hans@archlinux meltdown-chess-engine]$ ./docker/run.sh

# rebuild from container
[user@b11f50ca269a meltdown-chess-engine]$ ./scripts/build.sh --release

# build should work now
```

> Again, i managed to compile with Meson (v1.8.0) by adding "-13" to the GCC lines in ./targets/*.txt .. although you discouraged this 'hack', this seems most easy and logical to me.

This is not really a solution. It causes a whole lot more issues that it "solves" :) 
The compilation targets are supposed to stay generic and avoid having fixed compiler versions etc. defined.
The docker container _should_ fix compatibility issues. Just me who fk'ed it up, sorry :)

hansbinderup:
> @hansbinderup i wonder what you mean by that

When mounting a docker container the paths are changed, so the meson setup was effectively invalid.
I decided to push a simple fix for this: https://github.com/hansbinderup/meltdown-chess-engine/pull/128

>  .. other packages using Meson just work for me, i even created my own Meson files, eg. to compile older C code lacking a Makefile - i use a config file to set [binaries] and more. Here the user should set / adjust their GCC version. Isn't that "the Meson way" ? 

The reason you have an issue with compiling Meltdown is due to having a default older compiler.
This is where docker is the simplest and most efficient way to provide a standalone image with all tools installed :)
As a developer I can't guarantee that consumers have the correct versions installed, so the easiest and safest, is to provide this as a combined bundle.

> In meson.build we can also check the compiler version and echo a message / prompt the user how to continue the compilation. 

The build targets solve this issue in _most_ cases. And otherwise docker is possible (and preferred).

> Can Meson even download a proper (temp) GCC binary to compile - that would be nice !?

This is exactly what docker does :)

> It's good to suit all OS and users, but keep things simple.
> Docker may be nice but it used several Gb on my HD, afterwards i had to ask AI how to remove those Docker files (so, find the Docker folder), because i rarely use it.

I think this is strongly opinionated - docker is definitely the simplest way to handle compiler versions etc. We use it professionally all the time. 
The entire system plays nicely with CI, automated releasing etc.
Compilers, build systems etc also take up space and it's harder to delete than a container.

Another solution is to download the releases instead of compiling yourself :)

hansbinderup:
> > Another solution is to download the releases instead of compiling yourself :)
> 
> 
> 
> Newest asset, same problem 😀 >
> ```
> 
> $ ./meltdown-v1.6-linux-gcc-x86-64-v3-asset-B 
> 
> ./meltdown-v1.6-linux-gcc-x86-64-v3-asset-B: /lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.15' not found (required by ./meltdown-v1.6-linux-gcc-x86-64-v3-asset-B)
> 
> ```

1.6 doesn't have the static linkage fix :) 

It will be part of next release! Maybe I'll rebuild all releases for compatibility 👍

hansbinderup:
@tissatussa I've added a new release: https://github.com/hansbinderup/meltdown-chess-engine/releases/tag/v1.7
Releases should work for you know and going forward :)

@Hestehviskeren please don't be rude.

PR #124, opened 2025-06-02
searcher: create SearchTable, clean up picker state

clarashepherd:
Lighter move picker can be duplicated easily, so this commit also introduces multiple pickers, each with their own state.

We no longer expose the PV table etc. Instead, the search table now acts as an API for PV table, history and killer moves.

Bench 3078472.

https://openbench.bunny.beer/test/300/

### DIFF
--- a/src/search/move_picker.h
+++ b/src/search/move_picker.h
@@ -4,9 +4,9 @@
 
 #include "evaluation/see_swap.h"
 #include "movegen/move_types.h"
-#include "search/history_moves.h"
-#include "search/killer_moves.h"
-#include "search/pv_table.h"
+
+#include "search/search_tables.h"
+
 #include <cstdint>
 
 namespace search {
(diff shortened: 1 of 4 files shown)


src/search/move_picker.h
```diff
@@ -32,124 +32,102 @@ enum PickerPhase {
...
-
-    constexpr const PVTable& pvTable() const
+    void setPhase(PickerPhase phase)
     {
-        return m_pvTable;
+        m_phase = phase;
     }
 
-    constexpr KillerMoves& killerMoves()
+    template<Player player> constexpr std::optional<movegen::Move> pickNextMove(SearchTable& searchTable, const BitBoard& board, movegen::ValidMoves& moves, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt)
```
hansbinderup, line 40: Nitpick: All other methods have board as first argument, maybe keep it that way

clarashepherd: Done

PR #127, opened 2025-06-04
move_picker: add counter moves term

clarashepherd:
See https://www.chessprogramming.org/Countermove_Heuristic for an explanation.

We only store one 64*64 array of counter moves, rather than one for each colour, because indexing by `[from][to]` is sufficiently unique to induce favourable move ordering.

Bench 2946602

https://openbench.bunny.beer/test/305/


src/search/move_picker.h
```diff
@@ -42,7 +43,7 @@ class MovePicker {
         m_phase = phase;
     }
 
-    template<Player player> constexpr std::optional<movegen::Move> pickNextMove(const BitBoard& board, SearchTables& searchTables, movegen::ValidMoves& moves, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt)
+    template<Player player> constexpr std::optional<movegen::Move> pickNextMove(const BitBoard& board, SearchTables& searchTables, movegen::ValidMoves& moves, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt, std::optional<movegen::Move> prevMove = std::nullopt)
```
hansbinderup, line 46: I think this method is becoming a little bloated. Wouldn't it be easier to just provided the tables, moves etc to the constructor and then let this method be a little lighter? :)

src/search/search_tables.h
```diff
@@ -109,10 +111,21 @@ class SearchTables {
         m_historyMoves.update(board, move, ply);
     }
 
+    inline movegen::Move getCounterMove(const movegen::Move& prevMove) const
```
hansbinderup, line 114: copy is cheaper (16 bits)
```suggestion
    inline movegen::Move getCounterMove(movegen::Move prevMove) const
```

src/search/search_tables.h
```diff
@@ -109,10 +111,21 @@ class SearchTables {
...
+    {
+        return m_counterMoves.get(prevMove);
+    }
+
+    inline void updateCounterMoves(const movegen::Move& prevMove, const movegen::Move& counterMove)
```
hansbinderup, line 119: copy is cheaper (16 bits)
```suggestion
    inline void updateCounterMoves(movegen::Move prevMove, movegen::Move counterMove)
```

src/search/searcher.h
```diff
@@ -188,10 +188,12 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
 
         movegen::ValidMoves captures;
         core::getAllMoves<movegen::MoveCapture>(board, captures);
+        const auto prevMove = m_stackItr == m_stack.begin() ? std::nullopt : std::optional<movegen::Move>((m_stackItr - 1)->move);
```
hansbinderup, line 191: `std::make_optional` can auto deduce type
```suggestion
        const auto prevMove = m_stackItr == m_stack.begin() ? std::nullopt : std::make_optional((m_stackItr - 1)->move);
```

src/search/searcher.h
```diff
@@ -188,10 +188,12 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
...
+
         if (captures.count()) {
             fmt::print("Captures[{}]: ", captures.count());
 
-            while (const auto& moveOpt = picker.pickNextMove(board, m_searchTables, captures, m_ply)) {
+            while (const auto& moveOpt = picker.pickNextMove(board, m_searchTables, captures, m_ply, std::nullopt, prevMove)) {
```
hansbinderup, line 196: Does it make sense to use here?

src/search/searcher.h
```diff
@@ -343,18 +347,21 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
...
 
-        while (const auto moveOpt = picker.pickNextMove(board, m_searchTables, moves, m_ply, ttMove)) {
+        while (const auto moveOpt = picker.pickNextMove(board, m_searchTables, moves, m_ply, ttMove, prevMove)) {
             const auto move = moveOpt.value();
 
             if (!makeMove(board, move)) {
                 continue;
             }
 
+            m_stackItr->move = move;
```
hansbinderup, line 363: should probably go into `makeMove` where the other stack values are set

src/search/searcher.h
```diff
@@ -405,7 +412,14 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
 
             if (score >= beta) {
                 core::TranspositionTable::writeEntry(m_stackItr->hash, score, m_stackItr->eval, move, depth, m_ply, core::TtBeta);
+
                 m_searchTables.updateKillerMoves(move, m_ply);
+
+                if (m_stackItr != m_stack.begin()) {
```
hansbinderup, line 418: for these checks it might be "prettier" to simply use `!isRoot` (what the rest of our checks do)

src/search/counter_moves.h
```diff
@@ -0,0 +1,34 @@
...
+
+    movegen::Move get(const movegen::Move& prevMove) const
+    {
+        return m_counterMoves.at(prevMove.fromPos()).at(prevMove.toPos());
+    }
+
+    void reset()
+    {
+        for (auto& row : m_counterMoves)
+            std::ranges::fill(row, movegen::nullMove());
```
hansbinderup, line 27: if you add a type eg: `using CounterEntry = std::array<movegen::Move, s_amountSquares>` you can prettify `m_counterMoves` type and avoid iterating here:
```suggestion
        std::ranges::fill(m_counterMoves, CounterEntry {});
```

src/search/counter_moves.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include "movegen/move_types.h"
+#include <algorithm>
+
+namespace search {
+
+class CounterMoves {
+public:
+    void update(const movegen::Move& prevMove, const movegen::Move& counterMove)
```
hansbinderup, line 10: copy instead of ref

src/search/counter_moves.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /* only quiet moves */
+        if (counterMove.isCapture()) {
+            return; // nothing to do
+        }
+        m_counterMoves.at(prevMove.fromPos()).at(prevMove.toPos()) = counterMove;
+    }
+
+    movegen::Move get(const movegen::Move& prevMove) const
```
hansbinderup, line 19: copy instead of ref

hansbinderup (commented):
A few minor comments, otherwise looks good :+1: 
Nice to see the gains! Been a while since we added ordering heuristics :rocket:

src/search/searcher.h
clarashepherd: I'm not sure I agree. We only use the move for the counter moves, which are only updated inside negamax. So for qsearch we'd be adding moves to the stack and removing them, while knowing they'd never currently be used.

hansbinderup: > I'm not sure I agree. We only use the move for the counter moves, which are only updated inside negamax. So for qsearch we'd be adding moves to the stack and removing them, while knowing they'd never currently be used.

we never remove stack information. We just decrement the iterator. The memory is already allocated, so there's no overhead, as we always update the stack when incrementing the iterator.

I think it's error prone to update the stackItr in multiple locations. The only reason we don't update the `static_eval` is because it's extremely expensive if we were not to use it. But copying 16bits is literally free :)

src/search/counter_moves.h
```diff
@@ -0,0 +1,34 @@
...
+    void update(movegen::Move prevMove, movegen::Move counterMove)
+    {
+        /* only quiet moves */
+        if (counterMove.isCapture()) {
+            return; // nothing to do
+        }
+        m_counterMoves.at(prevMove.fromPos()).at(prevMove.toPos()) = counterMove;
+    }
+
+    movegen::Move get(movegen::Move prevMove) const
```
hansbinderup, line 19:
```suggestion
    inline movegen::Move get(movegen::Move prevMove) const
```

src/search/move_picker.h
```diff
@@ -144,12 +159,12 @@ class MovePicker {
 
     // Helper: calling inside loops will mean redundant colour checks
     constexpr std::optional<movegen::Move>
-    pickNextMove(const BitBoard& board, SearchTables& searchTables, movegen::ValidMoves& moves, uint8_t ply, const std::optional<movegen::Move>& ttMove = std::nullopt)
+    pickNextMove(const BitBoard& board, movegen::ValidMoves& moves)
```
hansbinderup, line 162: Nitpick: Line break (was there before)

PR #135, opened 2025-06-06
move_picker: move generation to picker

clarashepherd:
Move generation is currently happening in the searcher, but its only use is to be consumed by the picker.

This commit move the generation to the picker, which also will enable optimizations, e.g. staged move generation.

Bench 2389574

### DIFF
--- a/src/search/move_picker.h
+++ b/src/search/move_picker.h
@@ -1,11 +1,13 @@
 #pragma once
 
 #include "core/bit_board.h"
+#include "core/move_handling.h"
 
 #include "evaluation/see_swap.h"
 #include "movegen/move_types.h"
 
 #include "search/search_tables.h"
+#include "syzygy/syzygy.h"
 
 #include <cstdint>
 
@@ -17,7 +19,9 @@ enum KillerMoveType {
 };
 
 enum PickerPhase {
+    GenerateSyzygyMoves,
     Syzygy,
+    GenerateMoves,
     TtMove,
     PvMove,
     CaptureGood,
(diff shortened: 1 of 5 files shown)


src/search/move_picker.h
```diff
@@ -33,121 +34,129 @@ enum PickerPhase {
 
 class MovePicker {
 public:
-    MovePicker(
-        SearchTables& searchTables, uint8_t ply, std::optional<movegen::Move> ttMove = std::nullopt, std::optional<movegen::Move> prevMove = std::nullopt, PickerPhase phase = PickerPhase::TtMove)
+    MovePicker(SearchTables& searchTables, uint8_t ply, PickerPhase phase, movegen::MoveType moveType, std::optional<movegen::Move> ttMove = std::nullopt, std::optional<movegen::Move> prevMove = std::nullopt, std::optional<movegen::ValidMoves> syzygyMoves = std::nullopt)
```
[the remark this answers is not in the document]
clarashepherd: That's very helpful to know. I've moved the syzygy movegen to the picker :)

src/search/move_picker.h
```diff
@@ -342,15 +369,20 @@ class MovePicker {
...
+
     uint8_t m_ply {};
+    PickerPhase m_phase { PickerPhase::TtMove };
+    movegen::MoveType m_moveType { movegen::MovePseudoLegal };
     std::optional<movegen::Move> m_ttMove { std::nullopt };
     std::optional<movegen::Move> m_prevMove { std::nullopt };
-    PickerPhase m_phase { PickerPhase::TtMove };
+
+    movegen::ValidMoves m_moves {};
+    bool m_hasGeneratedMoves { false };
```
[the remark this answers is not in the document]
clarashepherd: I've changed `GenerateMoves` to a state - I like this recurring design theme of flags -> states - but for the syzygy generation I don't think it's so simple.

Currently in the searcher there's an early return, before the picking, depending on whether we have syzygy moves. To keep that in place, I added a `generateSyzygyMoves` function to the picker, rather than have it as a state.

I think this makes sense, because the `GenerateMoves` state was introduced to avoid multiple movegen calls. With syzygy we don't have the same issue - the seacher calls to generate, only once.

What do you think?

src/search/move_picker.h
```diff
@@ -342,15 +396,27 @@ class MovePicker {
...
     uint8_t m_ply {};
+    PickerPhase m_phase { PickerPhase::TtMove };
     std::optional<movegen::Move> m_ttMove { std::nullopt };
     std::optional<movegen::Move> m_prevMove { std::nullopt };
-    PickerPhase m_phase { PickerPhase::TtMove };
+
+    movegen::ValidMoves m_moves {};
+
+    static syzygy::WdlResult s_wdl;
+    static uint8_t s_dtz;
```
hansbinderup, line 414: if these are static then we should probably store them in the syzygy implementation. Doesn't feel right to have them in the move picker :)

clarashepherd: Moved this over :) Going forwards, I think a cleaner solution would be to convert the syzygy namespace to a class, and have a static object that owns the result data. Follow up?

src/search/move_picker.h
```diff
@@ -342,15 +396,27 @@ class MovePicker {
...
 };
+
+template<movegen::MoveType moveType>
+syzygy::WdlResult MovePicker<moveType>::s_wdl { syzygy::WdlResultTableNotActive };
+
+template<movegen::MoveType moveType>
+uint8_t MovePicker<moveType>::s_dtz {};
```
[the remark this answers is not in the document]
clarashepherd: Now redundant here, but also noted :) Thanks

src/search/searcher.h
```diff
@@ -331,36 +333,25 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
...
 
-            if (m_searchTables.isPvFollowing()) {
-                m_searchTables.updatePvScoring(moves, m_ply);
-            }
+            m_tbHits++;
+            return score;
         }
 
+        /* generateSyzygyMoves is not thread safe - allow primary searcher only to take this path! */
+        const auto phase = isSyzygyActive && isRoot && m_isPrimary ? PickerPhase::GenerateSyzygyMoves : PickerPhase::GenerateMoves;
```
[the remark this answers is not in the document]
clarashepherd: I was trying to avoid non-const definitions, but we've seen the elo penalty for the additional branching. I've implemented your suggestion :)

src/search/searcher.h
```diff
@@ -92,8 +92,9 @@ class Searcher : public std::enable_shared_from_this<Searcher> {
         assert(m_stackItr == m_stack.begin());
 
         m_searchTables.setPvIsFollowing(true);
-        m_wdl = syzygy::WdlResultTableNotActive;
-        m_dtz = 0;
+
+        if (m_isPrimary)
+            syzygy::resetResults();
```
hansbinderup, line 97: Maybe just move this to `reset` where we used to reset `m_wdl` and `m_dtz` :)

src/syzygy/syzygy.h
```diff
@@ -20,6 +20,15 @@ enum WdlResult {
 
 namespace {
 
+uint8_t s_dtz {};
+WdlResult s_wdl { WdlResultTableNotActive };
```
hansbinderup, line 24: Declared in header, so just to be safe.

```suggestion
static inline uint8_t s_dtz {};
static inline WdlResult s_wdl { WdlResultTableNotActive };
```

src/syzygy/syzygy.h
```diff
@@ -20,6 +20,15 @@ enum WdlResult {
...
+
+inline void resetResults()
```
hansbinderup, line 26: Reset is probably enough 
```suggestion
inline void reset()
```

src/search/move_picker.h
```diff
@@ -31,123 +35,147 @@ enum PickerPhase {
...
         , m_ply(ply)
+        , m_phase(phase)
         , m_ttMove(ttMove)
         , m_prevMove(prevMove)
-        , m_phase(phase)
     {
     }
 
-    void setPhase(PickerPhase phase)
+    constexpr void resetPicking(PickerPhase phase)
```
hansbinderup, line 50: Seems to be unused (remove if so)

src/search/move_picker.h
```diff
@@ -31,123 +35,147 @@ enum PickerPhase {
...
-    template<Player player> constexpr std::optional<movegen::Move> pickNextMove(const BitBoard& board, movegen::ValidMoves& moves)
+    template<Player player> constexpr std::optional<movegen::Move> pickNextMove(const BitBoard& board)
     {
         switch (m_phase) {
+        case GenerateSyzygyMoves: {
+            if (syzygy::generateSyzygyMoves(board, m_moves)) {
+                m_phase = PickerPhase::Syzygy;
+            } else {
+                m_phase = PickerPhase::GenerateMoves;
+            }
```
hansbinderup, line 69: nitpick: I personally prefer explicit bools. If you don't agree just leave as is :)
```suggestion
            const bool syzygyActive = syzygy::generateSyzygyMoves(board, m_moves);
            if (syzygyActive) {
                m_phase = PickerPhase::Syzygy;
            } else {
                m_phase = PickerPhase::GenerateMoves;
            }
```

hansbinderup (approved):
Looks good! Let's get it rolling :rocket:

src/syzygy/syzygy.h
clarashepherd: Done :)

PR #153, opened 2025-06-16
transposition: add tt pv | remove pv move picking

hansbinderup:

```
Elo   | -0.11 +- 2.66 (95%)
SPRT  | 8.0+0.08s Threads=1 Hash=128MB
LLR   | 2.96 (-2.94, 2.94) [-5.00, 0.00]
Games | N: 30612 W: 7549 L: 7559 D: 15504
Penta | [874, 3673, 6182, 3743, 834]
https://openbench.bunny.beer/test/396/
```

### DIFF
--- a/src/core/transposition.h
+++ b/src/core/transposition.h
@@ -15,9 +15,40 @@ enum TtFlag : uint8_t {
     TtBeta = 2,
 };
 
+/* Storage for smaller (< 8 bits) types stored in TT data
+ * 0b00000011 -> flag (2 bits)
+ * 0b00000100 ->   pv (1 bit) */
+struct TtInfo {
+    TtInfo() = default;
+    TtInfo(TtFlag flag, bool isPv)
+        : data((flag & s_flagMask)
+              | (isPv << s_pvShift))
+    {
+    }
+
+    /* tt flag - which bound the entry has */
+    inline TtFlag flag() const
+    {
+        return static_cast<TtFlag>(data & s_flagMask);
+    }
+
+    /* flag to indicate if this entry is/was following the PV line */
+    inline bool pv() const
+    {
+        return data & s_pvFlag;
+    }
+
+private:
+    uint8_t data;
+
+    static constexpr inline uint8_t s_flagMask { 0b11 };
+    static constexpr inline uint8_t s_pvFlag { 0b100 };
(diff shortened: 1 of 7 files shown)


hansbinderup:
@clarashepherd this should help with lazy move picking - no more dependencies to the PV scoring/following :rocket:

src/core/transposition.h
```diff
@@ -15,9 +15,40 @@ enum TtFlag : uint8_t {
...
+    }
+
+    /* tt flag - which bound the entry has */
+    inline TtFlag flag() const
+    {
+        return static_cast<TtFlag>(data & s_flagMask);
+    }
+
+    /* flag to indicate if this entry is following the PV line */
+    inline bool ttPv() const
```
clarashepherd, line 36: Do you think there's maybe a better name for this? Perhaps `isPv` or similar?

src/core/transposition.h
```diff
@@ -140,8 +174,8 @@ class TranspositionTable {
         /* only update entry if a better one is found */
         if (!sameKey
             || entryData.move.isNull()
-            || depth >= entryData.depth
-            || (flag == TtExact && entryData.flag != TtExact)) {
+            || depth + (2 * ttPv) >= entryData.depth
```
clarashepherd, line 177: Should this be a multiplication or a bitshift?

src/core/transposition.h
hansbinderup: I was a little afraid of calling it that as we already have a "isPv" flag which I don't want it to be mixed up with.
I can simply call it `pv` here (where it's bound to TT) and use `ttPv` in the searcher to make it explicit there? :)

src/core/transposition.h
hansbinderup: I did some initial research and I didn't find any engine that didn't use 2 here. 
It might actually be interesting to add this as a SPSA parameter instead and let the tuner have a go at it! But maybe we should leave that to a tuning build (hoping to do that after 1.8) :)

clarashepherd: Sounds good, I think that could be an interesting follow-up - mostly because I'm curious to know what the 'optimal' value might be here :)

clarashepherd (approved):

PR #154, opened 2025-06-17
searcher: isPv and isRoot as constexpr parameters

hansbinderup:
We previously had the "nullMove" check at compiletime. This is only used
in a single case why we could benefit from using `isPv` and `isRoot` at
compiletime instead.

These are used a few more times and will therefore gives us a slight
speedup.

NOTE: this commit also fixes the PV hack where we assumed that we were
following the PV line based on alpha and beta values.

Bench 1571664

```
Elo   | 2.83 +- 2.26 (95%)
LLR   | 2.96 (-2.94, 2.94) [0.00, 5.00]
Games | N: 37014 W: 9326 L: 9024 D: 18664
Penta | [801, 4134, 8416, 4274, 882]
https://openbench.bunny.beer/test/398/
```


clarashepherd (approved):
Nice marginal gains :)

PR #161, opened 2025-06-23
Clara/combine picker phases

clarashepherd:
This commit combines promotions and captures into a single phase.
Promotions therefore benefit from pre-computation during scoring phase.

Captures are sorted by their SEE value, good promotions are assigned the
value of a queen in the SEE swap, and bad promotions are scored highly
negatively. The latter two are inlined, in order to preserve constexpr
of the queen SEE score and build for SPSA.

Tactical moves have also been renamed to 'noisy'.

Bench 2953743

https://openbench.bunny.beer/test/487/

```
Elo   | 1.48 +- 4.15 (95%)
LLR   | -3.06 (-2.94, 2.94) [0.00, 10.00]
```

### DIFF
--- a/src/search/move_picker.h
+++ b/src/search/move_picker.h
@@ -23,15 +23,13 @@ enum PickerPhase {
     Syzygy,
     GenerateMoves,
     TtMove,
-    GenerateCaptureScores,
-    CaptureGood,
-    PromotionGood,
+    GenerateNoisyScores,
+    NoisyGood,
     KillerMoveFirst,
     KillerMoveSecond,
     CounterMove,
     HistoryMove,
-    PromotionBad,
-    BadCapture,
+    NoisyBad,
     Done,
 };
 
@@ -45,6 +43,10 @@ public:
         , m_ttMove(ttMove)
         , m_prevMove(prevMove)
     {
+        if constexpr (moveType == movegen::MoveCapture) {
+            if (m_ttMove && !m_ttMove->isCapture())
+                m_ttMove.reset();
+        }
     }
 
     constexpr uint16_t numGeneratedMoves()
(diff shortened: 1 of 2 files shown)


src/search/move_picker.h
```diff
@@ -18,20 +18,23 @@ enum KillerMoveType {
...
 enum PickerPhase {
     GenerateSyzygyMoves,
     Syzygy,
     GenerateMoves,
     TtMove,
-    GenerateCaptureScores,
-    CaptureGood,
-    PromotionGood,
+    GenerateTacticalScores,
+    TacticalGood,
```
hansbinderup, line 32: The rest of the code base uses "noisy" instead of tactical :)

src/search/move_picker.h
```diff
@@ -229,25 +214,29 @@ class MovePicker {
...
-    void generateCaptureScores(const BitBoard& board)
+    void generateTacticalScores(const BitBoard& board)
     {
 
         for (uint16_t i = 0; i < m_moves.count(); i++) {
-            if (m_moves[i].isCapture()) {
+            if (m_moves[i].isCapture())
                 m_scores[i] = evaluation::SeeSwap::run(board, m_moves[i]);
-            }
+            // Capture-promotions are handled in SEE swap
```
hansbinderup, line 223: Nitpick: don't like comments between if cases when they're not scoped (can easily lead to annoying syntax errors)

hansbinderup (commented):
A few initial comments :)

src/search/move_picker.h
```diff
@@ -18,20 +18,23 @@ enum KillerMoveType {
     Second,
 };
 
+enum PromotionOffsets : int32_t {
+    Good = 8000,
```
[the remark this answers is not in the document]
clarashepherd: Good to know :) Now redundant, but I'll keep this in mind going forwards.

src/search/move_picker.h
clarashepherd: Changed :)

src/search/move_picker.h
clarashepherd: Good to know :) Now redundant, but I'll keep this in mind going forwards.

PR #162, opened 2025-06-25
evaluation: add king pawn cache

hansbinderup:

Also sorry for the big diff.. But it required quite a few refactors to get working.. I've tried to divide the major parts into, still, bigger commits. Should be easier to review going through one commit at a time :)

```
Elo   | 21.83 +- 9.17 (95%)
LLR   | 3.03 (-2.94, 2.94) [0.00, 5.00]
Games | N: 2454 W: 692 L: 538 D: 1224
Penta | [54, 245, 485, 379, 64]
https://openbench.bunny.beer/test/440/
```


src/core/move_handling.h
```diff
@@ -103,16 +110,22 @@ constexpr void performPromotionMove(BitBoard& newBoard, movegen::Move move)
...
+
+            if (victim == WhitePawn || victim == BlackPawn) {
+                core::hashPiece(*victim, move.toPos(), newBoard.kpHash);
+            }
         }
     }
 
     switch (move.promotionType()) {
     case PromotionNone:
+        assert(false);
```
clarashepherd, line 128: I think we're currently mixing hard and soft error handling throughout the codebase.

e.g. `getAttackerAtSquare` fails with a bad optional access, rather than assert we've added.

Do you think it's worth giving a moment's thought towards standardizing our error handling in a lightweight way?

src/evaluation/static_evaluation.h
```diff
@@ -22,45 +23,86 @@ namespace evaluation {
...
+    TermScore fetchOrStoreKpCache(const BitBoard& board, TermContext& ctx)
+    {
+        TermScore score(0, 0);
+
+        const auto kpProbe = m_kpCache.probe(board.kpHash);
+        if (kpProbe.has_value()) {
+            score += kpProbe->score;
+            ctx.passedPawns = kpProbe->passedPawns;
+        } else {
+            APPLY_SCORE(getStaticKingPawnScore, board, ctx);
```
clarashepherd, line 98: Could you please explain what's happening for a cache miss?

src/core/move_handling.h
hansbinderup: Assertions are free (no runtime penalty). Ofc only for release builds - which is what we optimize for.

We shouldn't overdo it, but there's absolutely no harm in adding assertions like these! 

Runtime error handling where not optimized away doing release builds are no go (maybe what you're refering to). It's simply too expensive. 

Your example with 'std::optional' has implicit assertion so it's the same thing :)

src/evaluation/static_evaluation.h
hansbinderup: We need to populate the cache before we can retrieve it.
So either we just bite the apple and do as before - or maybe we're able to fetch a cache. Best case it's faster, worst case it's same as before this PR :)

src/core/board_defs.h
```diff
@@ -153,3 +153,5 @@ constexpr static inline size_t s_maxThreads { 128 };
 
 using TimePoint = std::chrono::time_point<std::chrono::steady_clock>;
 constexpr static inline std::chrono::milliseconds s_defaultMoveOverhead { 50 };
+
+constexpr static inline std::string_view s_startPosFen { "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1" };
```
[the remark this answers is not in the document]
hansbinderup: Reset no longer works as there's a recursive dependency to zobrist and BitBoard. 
So I thought the easiest and cleanest way was to just remove the method and let Fen handle it for us :)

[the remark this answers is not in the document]
hansbinderup: reset was also a bit off. It's just the start position - not an actual reset so now it's a bit more explicit :)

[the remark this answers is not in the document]
clarashepherd: Makes sense. I also like how it's more concise and aligned with the rest of the codebase :)

src/evaluation/static_evaluation.h
hansbinderup: I've added more information to the commit message as well.
But the theory is that the king pawn structure is more or less static. There's only so few possibilities (pawns can only move one step and only one direction).
So previously the pawn evaluation was more or less static - I've prepared for this already by adding more dynamic or volatile terms as non-piece terms: eg passed pawns where we need the full context to evaluate.

But for these "static terms" it's a nice and relatively simple way to get some speed ups. Also the cache is dum as can be - but it's still very efficient which just proves the point :)

PR #175, opened 2025-07-07
move_picker: pre-score quiet phases

clarashepherd:
Slightly increased node count, because history heuristic updates during search. Net elo gain due to the speedup of pre-calculating scores.

Bench 3271931

```
Elo   | 14.32 +- 9.82 (95%)
LLR   | 2.30 (-2.94, 2.94) [-5.00, 0.00]
```

https://openbench.bunny.beer/test/504/

### DIFF
--- a/src/search/move_picker.h
+++ b/src/search/move_picker.h
@@ -18,6 +18,12 @@ enum KillerMoveType {
     Second,
 };
 
+enum QuietMoveOffsets {
+    KillerMoveFirst = 100003,
+    KillerMoveSecond = 100002,
+    CounterMove = 100001,
+};
+
 enum PickerPhase {
     GenerateSyzygyMoves,
     Syzygy,
@@ -25,10 +31,8 @@ enum PickerPhase {
     TtMove,
     GenerateNoisyScores,
     NoisyGood,
-    KillerMoveFirst,
-    KillerMoveSecond,
-    CounterMove,
-    HistoryMove,
+    GenerateQuietScores,
+    QuietMove,
     NoisyBad,
     Done,
 };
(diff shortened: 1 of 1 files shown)


src/search/move_picker.h
```diff
@@ -18,17 +18,21 @@ enum KillerMoveType {
     Second,
 };
 
+enum QuietMoveOffsets {
+    KillerMoveFirst = 30003,
+    KillerMoveSecond = 30002,
+    CounterMove = 30001,
```
hansbinderup, line 24: what are these values based on? Maybe scale them excessively so we guarantee that we don't overflow history scores :)

src/search/move_picker.h
```diff
@@ -252,68 +237,54 @@ class MovePicker {
...
-        for (uint16_t i = 0; i < m_moves.count(); i++) {
-            if (!m_moves[i].isNull() && m_moves[i].isQuietMove() && m_moves[i] == counterMove) {
-                auto pickedMove = m_moves[i];
-
-                m_moves.nullifyMove(i);
-
-                return pickedMove;
+                // Debug
+                if (m_scores[i] > QuietMoveOffsets::CounterMove)
+                    assert(false);
```
hansbinderup, line 262: `assert` is supposed to be used like this:
```suggestion
                assert(m_scores[i] < QuietMoveOffsets::CounterMove);
```

then when compiling with `NDEBUG` the entire macro results in `void(0)`

src/search/move_picker.h
```diff
@@ -252,68 +237,54 @@ class MovePicker {
...
             }
         }
-        return std::nullopt;
     }
 
-    template<Player player>
-    constexpr std::optional<movegen::Move> pickHistoryMove(const BitBoard& board)
+    constexpr std::optional<movegen::Move> pickQuietMove()
     {
+        std::optional<movegen::Move> bestMove { std::nullopt };
```
hansbinderup, line 269: nitpick: is already default
```suggestion
        std::optional<movegen::Move> bestMove { };
```

hansbinderup (approved):
Just a few minor comments, otherwise lgtm 💯

src/search/move_picker.h
clarashepherd: I used large values that were far away from history move scores, to force ordering.

Do you have any suggestions for alternative values? I'm not sure how we'd have an overflow here.

src/search/move_picker.h
clarashepherd: That's really hepful to know know about `NDEBUG` - thanks!

On reflection, I don't think this statement should be included in either debug or release builds - because it doesn't correspond to a genuine bug, only sub-optimal choices of move ordering. I'll remove it.

src/search/move_picker.h
```diff
@@ -252,68 +237,54 @@ class MovePicker {
...
-        const auto counterMove = m_searchTables.getCounterMove(m_prevMove.value());
-
-        for (uint16_t i = 0; i < m_moves.count(); i++) {
-            if (!m_moves[i].isNull() && m_moves[i].isQuietMove() && m_moves[i] == counterMove) {
-                auto pickedMove = m_moves[i];
-
-                m_moves.nullifyMove(i);
-
-                return pickedMove;
+                // Debug
```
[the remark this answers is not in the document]
clarashepherd: Removed, thanks :)

src/search/move_picker.h
clarashepherd: Changed, but the now-removed debug statement already confirmed that history scores were always lower than the offsets.

Also started a new test just to ensure nothing's broken.

hansbinderup: > Changed, but the now-removed debug statement already confirmed that history scores were always lower than the offsets.

*Always* is a big word. Maybe during bench, but that's also not a deep search. 
The history scores will keep increasing during a search, so it's better to be safe than sorry :)

PR #179, opened 2025-07-11
transposition: use multiply-shift hashing for index

hansbinderup:
Use multiply-shift hashing instead of modulo for faster index computation

This improves performance over `%` at the cost of slight bias when the table size is not a power of two.

NOTE: this works as long as the table is smaller than 2^64 which should never happen in real life x) (our tables are limited from UCI to an actual realistic value)

Bench 2881777

```
Elo   | 10.77 +- 6.43 (95%)
LLR   | 2.95 (-2.94, 2.94) [0.00, 5.00]
Games | N: 5582 W: 1534 L: 1361 D: 2687
Penta | [149, 659, 1049, 738, 196]
https://openbench.bunny.beer/test/533/
```

### DIFF
--- a/src/core/transposition.h
+++ b/src/core/transposition.h
@@ -109,7 +109,7 @@ public:
     constexpr static void prefetch([[maybe_unused]] uint64_t key)
     {
         assert(s_tableSize > 0);
-        __builtin_prefetch(&s_table[key % s_tableSize]);
+        __builtin_prefetch(&s_table[computeHashIndex(key)]);
     }
 
     static std::size_t getSizeMb()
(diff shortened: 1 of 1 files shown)


src/core/transposition.h
```diff
@@ -109,7 +109,7 @@ class TranspositionTable {
     constexpr static void prefetch([[maybe_unused]] uint64_t key)
     {
         assert(s_tableSize > 0);
-        __builtin_prefetch(&s_table[key % s_tableSize]);
```
clarashepherd, line 112: To avoid statistical bias, at line 97 why don't we just make `s_tableSize` a power of two? :)

hansbinderup: because it's scaled based on UCI input. We can't guarantee that the user always inputs a power of two size.
We can ofc scale it to a power of two, but we might miss out on potentially huge amounts of TT storage instead which will most likely cause a bigger loss than win (eg downscale to 128 if eg 200 was used).

Most tournaments are aware of this and always use power of 2 sizes :)

PR #180, opened 2025-07-12
searcher: add extended futility pruning

hansbinderup:
This commit adds initial static forward pruning.

The first pruning technique is called "extended futility pruning" (EFP). The theory here is that after searching a certain amount of nodes, if the evaluation is marginally worse than alpha then we can assume that we're not going to improve that with "bad moves".

The "bad moves" are currently filtering out quiets and bad promotions. We could be more aggresive and also filter bad captures, but let's keep it a bit safer for now.

This scheme also makes it easy to apply more types of static forward pruning.

Bench 2820965

```
Elo   | 11.16 +- 6.56 (95%)
LLR   | 3.01 (-2.94, 2.94) [0.00, 5.00]
Games | N: 5606 W: 1583 L: 1403 D: 2620
Penta | [169, 642, 1052, 720, 220]
https://openbench.bunny.beer/test/538/
```

### DIFF
--- a/src/search/move_picker.h
+++ b/src/search/move_picker.h
@@ -18,10 +18,11 @@ enum KillerMoveType {
     Second,
 };
 
-enum QuietMoveOffsets {
+enum MovePickerOffsets : int32_t {
     KillerMoveFirst = 100003,
     KillerMoveSecond = 100002,
     CounterMove = 100001,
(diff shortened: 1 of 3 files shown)


clarashepherd (approved):
Four more tunable parameters, let's go that elo :medal_sports:

PR #192, opened 2025-07-23
move_handling: improve castle rights | mask_tables: move tables and clean up mask table usage

hansbinderup:

Mostly a clean up but also a slight performance gain in terms of the castle lookup.

```
Elo   | 3.23 +- 4.65 (95%)
LLR   | 2.95 (-2.94, 2.94) [-5.00, 0.00]
Games | N: 9788 W: 2399 L: 2308 D: 5081
Penta | [248, 1171, 1990, 1212, 273]
https://openbench.bunny.beer/test/569/
```


clarashepherd (approved):
Looks good! My only suggestion is that perhaps the second and third commits could be combined, because they both perform a cleanup and reduce redundancy :)

The colour-swap trick is neat :zap:

hansbinderup:
> Looks good! My only suggestion is that perhaps the second and third commits could be combined, because they both perform a cleanup and reduce redundancy :)

You're right, I'll squash them :+1: 

> The colour-swap trick is neat ⚡

Glad you like it :rocket:

PR #194, opened 2025-07-29
see_swap: fix initial and promotion score

hansbinderup:
Previously we had two issues:

1. we returned 0 when no counter attackers. Here we should return the gain of no counter attacks.
2. Promotion score was off - we didn't add the gain of the promoted piece. As the SEE values are off by our tuning I've reset these to primitive values. They'll be adjusted in next spsa tuning.

Bench 1410566

```
Elo   | 5.17 +- 3.90 (95%)
LLR   | 2.98 (-2.94, 2.94) [0.00, 5.00]
Games | N: 14102 W: 3563 L: 3353 D: 7186
Penta | [362, 1657, 2853, 1767, 412]
https://openbench.bunny.beer/test/596/
```

### DIFF
--- a/src/evaluation/see_swap.h
+++ b/src/evaluation/see_swap.h
@@ -53,16 +53,13 @@ public:
 
         /* subtract pawn from promotion value */
         if (promotionPiece.has_value()) {
-            gain[depth] -= s_pieceValues[Pawn];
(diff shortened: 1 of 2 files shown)


clarashepherd (approved):
Nice to see this finally fixed up :100:

PR #195, opened 2025-07-31
searcher: align history moves with correction

hansbinderup:
When adding correction history we made a filter that ensure that the history we use is stable. Currently we add history moves for whatever cutoff. Instead improve this by filtering the history moves we use.

Bench 1383681

```
Elo   | 13.44 +- 4.77 (95%)
LLR   | 2.95 (-2.94, 2.94) [0.00, 2.00]
Games | N: 9594 W: 2539 L: 2168 D: 4887
Penta | [241, 1043, 1931, 1268, 314]
https://openbench.bunny.beer/test/601/
```


clarashepherd:
Best elo gains for moving one line so far?? :1st_place_medal:

hansbinderup:
> Best elo gains for moving one line so far?? 🥇

Wish that it was this easy every time.. :laughing:

PR #197, opened 2025-08-05
searcher: add see pruning

hansbinderup:
SEE pruning is an attempt at pruning branches by evaluating the worst/best case scenario in a simplified way.

If the SEE score is significantly low then we can assume that the given branch is not worth searching.

This implementation is very similar to our capture score but here we just compute a balance - and quiet moves are also accepted.

Bench 1209330

```
Elo   | 7.16 +- 3.40 (95%)
LLR   | 2.96 (-2.94, 2.94) [0.00, 2.00]
Games | N: 19080 W: 4874 L: 4481 D: 9725
Penta | [510, 2212, 3803, 2405, 610]
https://openbench.bunny.beer/test/613/
```

### DIFF
--- a/src/evaluation/see_swap.h
+++ b/src/evaluation/see_swap.h
@@ -17,7 +17,78 @@ namespace evaluation {
 
 class SeeSwap {
 public:
-    static inline int32_t run(const BitBoard& board, movegen::Move move)
+    static inline bool isGreaterThanMargin(const BitBoard& board, movegen::Move move, int32_t margin)
+    {
+        if (move.isCastleMove()) {
+            return margin >= 0;
+        }
+
+        /* as we're comparing with margin we can simplify handling by comparing to zero instead */
+        int32_t balance = -margin;
+
+        Player player = board.player;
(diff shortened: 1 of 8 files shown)


src/evaluation/see_swap.h
```diff
@@ -17,7 +17,77 @@ namespace evaluation {
...
+            const auto piece = getLeastValuableAttacker(board, attackers, occ, player);
+            if (!piece.has_value())
+                break; /* no more attackers */
+
+            /* illegal position if king is attacked after capturing */
+            if (utils::isKing(*piece) && ((attackers & occ & board.occupation[nextPlayer(board.player)]) != 0)) {
+                break;
+            }
+
+            if (player == board.player) {
```
clarashepherd, line 75: Is there some slicker syntax available here? This is branching with one-liners inside.

clarashepherd (approved):
A few nitpicks, otherwise looking great :) Nice to see the SEE swap getting an upgrade!

